PR #16, opened 2023-01-24
Updating PETSc Docker image tag.

### LINKED ISSUE #11: Update Docker image with Matt's vertical geometry fixes
When we start to deal with problems that involve differing elevations, we'll need some new PETSc code from the `knepley/fix-plex-geom-1d` branch. At the time of writing, we have plenty of work to do in planar (flat) geometry, so this isn't urgent, but it's worth tracking,.

jeff-cohere:
Closes #11


jeff-cohere:
Actually, it looks like this PETSc rev doesn't have Matt's fix (see #11).

Issue #12, opened 2023-01-24
Create a C driver program capable of running a flat dam-break problem


jeff-cohere:
As a first step toward running [this validation problem](https://rdycore.atlassian.net/wiki/spaces/PD/pages/15237121/Flow+Partial+2D+dam+break+problem), we need a driver program that accepts input specifying initial conditions and geometry for a dam through which water flows. We'll begin with flat geometry and no dam walls to demonstrate that we can do 2D flow.

The driver program will use [YAML input](https://rdycore.atlassian.net/wiki/spaces/PD/pages/24576001/RDycore+configuration+file) that specifies all relevant information (with options for overriding parameters from the command line).

PR #18, opened 2023-02-13
Adding C and Fortran drivers and a bunch of infrastructure

### LINKED ISSUE #15: Create a Fortran driver program
When #12 has been completed, we'll create a Fortran version of the C driver program used to run the dam-break problem. We should be able to have the C code read all relevant input data and set up the simulation in an analogous format.

jeff-cohere:
Sorry for the large PR. I'm pushing hard to get us off the ground and didn't want to pester everyone with a bunch of small details.

In this PR:
* I've implemented a YAML parser for our [configuration file spec](https://rdycore.atlassian.net/wiki/spaces/PD/pages/24576001/RDycore+configuration+file#Iteration-3).
* I've created some basic setup and run elements to support a standalone driver
* I've created an `rdycore` Fortran module that wraps these basic elements
* I've put together simple C and Fortran versions of our standalone driver
* I've added a driver test with an input file to replicate the `ex2b` simulation in the toy-problems repo
* I've added some GMsh geometries that are used to generate meshes for driver tests. There are quad and tri versions of this mesh.
* I've put together some basic logging utilities.

There's no simulation logic in there yet. That (#12) is next on my list. This at least gives us several things to discuss.

Closes #15


jeff-cohere:
Okay, it looks like the massively over-engineered/underexplained CMake code coverage logic I found on the Internets is not going to work. I'll replace that with a more reasonable implementation. In the meantime, any questions/comments/concerns are welcome.

bishtgautam:
I needed the following changes to get the code to compile on my ubuntu laptop.

```
>git diff
diff --git a/src/read_config_file.c b/src/read_config_file.c
index 84f5c28..3304dee 100644
--- a/src/read_config_file.c
+++ b/src/read_config_file.c
@@ -195,11 +195,14 @@ static PetscErrorCode CloseSection(YamlParserState *state, RDy rdy) {
     } else {                     // we just need to pop out of this subsection
       switch (state->section) {  // move up one section
         case PHYSICS_SEDIMENT_SECTION:
+          break;
         case PHYSICS_SALINITY_SECTION:
+          break;
         case PHYSICS_FLOW_SECTION:
           state->section = PHYSICS_SECTION;
           break;
         default:
+          break;
       }
     }
     state->inside_subsection = PETSC_FALSE;
@@ -904,6 +907,7 @@ static PetscErrorCode HandleYamlEvent(WhichPass pass, yaml_event_t *event, YamlP
           break;
         default:
           // we ignore everything else for now
+          break;
       }
     } else {  // second pass
       // in the second pass, we parse secions that refer to other sections
@@ -918,6 +922,7 @@ static PetscErrorCode HandleYamlEvent(WhichPass pass, yaml_event_t *event, YamlP
           PetscCall(ParseSources(event, state, rdy));
           break;
         default:
+          break;
       }
     }
   }
```

I'm having other struggles running code on my ubutu laptop and will follow up on them on slack.

jeff-cohere:
A few issues remain (in addition to whatever is living in your heads that you haven't yet shared):
* @bishtgautam is seeing an issue where PETSC_MPIEXEC includes the `--oversubscribe` flag, which confuses our testing machinery.
* The YAML parser needs more test coverage, since I implemented the entire spec and don't have unit tests for it. It shouldn't be too hard to get some of those in place.
* @bishtgautam and I have been discussing whether it's reasonable to require someone to have `gmsh` on their system in order to enable tests. I'm in favor of this, since it means we can create geometry files with variable resolution for convergence testing, but Gautam wants to minimize the number of things that people have to have on their systems.

knepley:
@jeff-cohere Configure puts --oversubscribe in if it is OpenMPI since the test performance is really bad without it. I could put something in to take it out if we want.

I lean toward not requiring things, including `gmsh`. We can manage uniform refinement internally. I agree that we cannot make the initial mesh without gmsh. I would probably check in both the script and the initial mesh.

jedbrown:
Configure adds `--oversubscribe` because otherwise `mpiexec` will refuse to run on more than the number of physical cores (i.e., not including hyperthreads).

It would be possible to have an archive for generated meshes. For example, we could have a job that adds generated `.msh` files to a repo or hidden branch. Maybe the `.msh` files are small enough, but I don't love the duplication (inevitably, someone modifies the `.geo` file and doesn't commit the new `.msh` file). Not a big deal though at this stage.

jedbrown:
TBH, I'd assumed we would use PetscOptions for parsing these input files (PETSc uses libyaml internally). If you run with `-options_file foo.yaml`, it populates the options database and you can query everything from your code with `PetscOptionsEnum(...)` and so on. I feel really bad that you wrote a ton of code calling libyaml directly, maybe because we didn't get a chance to talk through this? I'm concerned about the complexity of having things that are parsed manually in this way and other things that populate the options database, with non-obvious to users rules about what is an option for PETSc and what is an RDycore option. What do you think?

jeff-cohere:
> TBH, I'd assumed we would use PetscOptions for parsing these input files (PETSc uses libyaml internally). If you run with `-options_file foo.yaml`, it populates the options database and you can query everything from your code with `PetscOptionsEnum(...)` and so on. I feel really bad that you wrote a ton of code calling libyaml directly, maybe because we didn't get a chance to talk through this? I'm concerned about the complexity of having things that are parsed manually in this way and other things that populate the options database, with non-obvious to users rules about what is an option for PETSc and what is an RDycore option. What do you think?

Don't feel bad, Jed! I investigated PETSc's YAML parser after you pointed it out (and I even saw that you use libyaml internally, though with internal linkage). The reason I decided to do it "manually" is that it doesn't seem possible to traverse a set of mappings with names that you don't know beforehand (such as the names of conditions that we use in our input spec). I've used libyaml on other projects, and it wasn't particularly difficult to write this parser. Really, I was motivated by the need for a slightly more expressive way of expressing initial/boundary conditions and sources/sinks, which are always a little tricky. But I understand your concerns about the complexity of not just using PETSc's options database, which is very convenient and simple.

I wouldn't be crestfallen if we ended up using PETSc's YAML parser, but I wanted to demonstrate that we can support a fairly expressive input format without a ton of effort. If there's a way we can do it more simply that doesn't cost us too much clarity (since we're going to be using this repeatedly to specify boundary value problems), I'm all for it. As far as overriding options, I think we'll have to document how it's done no matter what, so that doesn't concern me too much.

knepley:
Oh, I would be interested in moving that sort of thing in the library. Jed and I already do something similar with boundary conditions with the normal syntax, where one options opens up other options with created names.

jedbrown:
I agree that this should go into PETSc. I know the current system is slightly redundant, but have found it generally useful to be able to have unused material descriptions in the yaml. There is already special handling for `-mg_levels_ksp_type` implying `-mg_levels_{n}_ksp_type` for each `n` (unless the more specific option is available). I think we can have a way to obtain an array of all `x` such that `-prefix_{x}_suffix` appears in the database. Would that satisfy your needs, or are you thinking of a different mechanism to handle yaml arrays?

jeff-cohere:
I would love to see PETSc support something like this, and I'd be happy to help if that made it happen faster.

Jed, if `x` could be an integer or a string, that would handle all the cases I'm interested in (and likely all cases in general could be mapped onto this feature). I'm thinking more of dictionary-like mappings than arrays, which fits nicely into YAML. In particular, Python's C API offers a [method for traversing dictionary key/value pairs](https://docs.python.org/3/c-api/dict.html#c.PyDict_Next) that could inform our discussion. Let's discuss elsewhere when it's appropriate.

For this PR, all the logic in question is isolated to a single source file, and a good deal of that source file is error checking and validation that we'll want to keep around. Switching over to PETSc when we're ready should be relatively easy.

Thanks for bringing this up! I'm grateful that you're entertaining my ideas.

include/private/rdycoreimpl.h
```diff
@@ -28,49 +129,126 @@ struct _p_RDy {
...
+
+  // mesh surfaces
+  PetscInt   num_surfaces;
+  PetscInt   surface_ids[1+MAX_SURFACE_ID];
+  RDySurface surfaces[1+MAX_SURFACE_ID];
+
+  // Table of named sets of flow, sediment, and salinity conditions. Used by
+  // initial/source/boundary conditions below.
+  PetscInt             num_flow_conditions;
+  RDyFlowCondition     flow_conditions[1+MAX_CONDITION_ID];
```
bishtgautam, line 190: Why not make this a pointer and allocate the appropriate size depending on the entries in the yaml file instead of using `MAX_CONDITION_ID`?

jeff-cohere: Good question. I did this because we read in these conditions one at a time because of the way input is parsed. In order to allocate the right number of conditions, we'd have to read everything into an intermediate representation beforehand, which I decided not to do because it's more code. I think this will work well enough to get us started. If we teach PETSc's options database to parse this kind of YAML, it will serve as that intermediate representation and then we can do what you propose.

bishtgautam: sounds good.

bishtgautam (approved):

donghuix (approved):

jeff-cohere:
Okay, I think this is ready to go even if it decreases our code coverage a little bit. 90% is pretty good.

jeff-cohere:
Any other thoughts on this PR before we merge?

bishtgautam:
Thanks @jeff-cohere for this massive PR. It is good to go and we have identified a few enhancements that will be added in upcoming PRs.

PR #24, opened 2023-03-01
Implemented file-based initial conditions (and default boundary edge assignment)

### LINKED ISSUE #23: Add support for specifying an IC for the whole domain
Presently, IC is specified for each region in the `.yaml` as:

```
initial_conditions:
  1: # upstream region
    flow: dam_top_ic
  2: # downstream region
    flow: dam_bottom_ic

# NOTE: We use reflecting boundary conditions for the boundary, top_wall, and
# NOTE: bottom_wall surfaces

flow_conditions:
  dam_top_ic:
    type: dirichlet
    height: 10
    momentum: [0, 0]
  dam_bottom_ic:
    type: dirichlet
    height: 5
    momentum: [0, 0]
```

Add the ability to specify an IC for the entire domain without the need to define regions via

```
initial_condition
  file: <petsc-binary-vector-in-natural-order.bin>
```

jeff-cohere:
This implements the alternative method of specifying initial conditions using a file. See [our specification](https://rdycore.atlassian.net/wiki/spaces/PD/pages/24576001/RDycore+configuration+file#Iteration-3) for details.

@bishtgautam , I know we discussed putting this into an `initial_condition` section instead of `initial_conditions`, but this would introduce the possibility of a user specifying both `initial_condition` and `initial_conditions` sections, which I think would be confusing.

In addition, we now mark boundary edges on a mesh and intersect all face sets with it to make sure every boundary edge is either explicitly assigned to a surface OR assigned to an additional "catch-all" surface. This seems to work properly, but I'd like to implement a test or two that stresses it. More on that below.

Closes #23


jeff-cohere:
@knepley and @jedbrown : what PETSc viewers can I use to write a DMPlex mesh to a file? I know Exodus is supported, but I'd like an alternative for non-NetCDF setups. I'm putting together some simple tests that create files containing meshes with and without face sets so we can make sure everything works properly.

src/rdysetup.c
```diff
@@ -202,88 +210,160 @@ static PetscErrorCode FindSalinityCondition(RDy rdy, const char *name, PetscInt
...
+      PetscCall(DMLabelGetStratumIS(label, region_id, &cell_is));
+      if (cell_is) {
+        RDyRegion *region  = &rdy->regions[r];
+        rdy->region_ids[r] = region_id;
+        ++r;
+
+        PetscInt num_cells;
+        PetscCall(ISGetLocalSize(cell_is, &num_cells));
+        if (num_cells > 0) {
+          rdy->region_ids[rdy->num_regions] = region_id;
```
bishtgautam, line 243: Should this be `rdy->region_ids[r-1] = region_id`? (`r-1` because `r` was incremented on [238L](https://github.com/RDycore/RDycore/pull/24/files#diff-75d4508176a97a2c171e581ffa0d15d8d268b032c241fe78fef679b7d55652aaR238)).

bishtgautam: On second thought, this line should be deleted as you have already set the value on 237L.

jeff-cohere: Good find. Actually, I think we can get rid of that line, because line 237 already does what this one intends.

bishtgautam (commented):
I'm wondering if we should edit the names of the following two structs: (1) `RDyRegion` and (2)`RDySurface` (I'm not loving the "_Surface_" term).
- Option 1: Rename them as `RDyRegionCell` and `RDyRegionEdge`
- Option 2: Create a unified struct `RDyRegion`

```
typedef struct {
  PetscInt   type; ! cell or edge
  PetscInt   *ids;
  PetscInt    num;
} RDyRegion;
```

bishtgautam:
We should also add a new yaml file the use `share/meshes/DamBreak_grid5x10.exo` and `share/initial_conditions/DamBreak_grid5x10_wetdownstream.IC` as done in the toy-problem repo at  
https://github.com/RDycore/toy-problems/blob/main/.github/workflows/auto_build.yml#L39

jeff-cohere:
Yeah, I agree that the `Surface` terminology is confusing in 2D. What about `Boundary` instead? Then we can talk about the domain boundary to mean the boundary of the entire problem, and other boundaries would separate regions. A "region of edges" doesn't seem very intuitive to me (though in this context its meaning is clear).

bishtgautam:
`RDyBoundary` introduces a new term _Boundary_ that doesn't directly map onto Cell/Edge/Vertices. Thus, I'm included to go with either of the two options that I defined in the [above comment](https://github.com/RDycore/RDycore/pull/24#pullrequestreview-1320692347).

jeff-cohere:
Well, a "surface/boundary" relates to edges in exactly the same way as a "region" relates to cells--an edge can belong to a surface/boundary, just like a cell can belong to a region. We need concepts for both of these things.

Further, a surface/boundary is not really a "type" of region--it is the difference between the closed set consisting of the region and its bounding edges, and the open set that is region itself.

The term "surface" is a bit confusing in the context of this project, because we usually think of a 2D surface embedded in a 3D space, and we're really discussing curves in our 2D space. The term "boundary", however, does seem more descriptive for the set of edges bounding a portion of the computational domain, so I'm inclined to defend it. But if this doesn't convince you that we should call it a "boundary", I'll change it to either of the options you've suggested.

jeff-cohere:
> We should also add a new yaml file the uses `share/meshes/DamBreak_grid5x10.exo` and `share/initial_conditions/DamBreak_grid5x10_wetdownstream.IC` as done in the toy-problem repo at https://github.com/RDycore/toy-problems/blob/main/.github/workflows/auto_build.yml#L39

We can do this, but it will force us to require NetCDF support in every installation in order to read the Exodus file. If that's okay, I'll add these files to the repo.

jedbrown:
Hi, @jeff-cohere. You can use `vtk:out.vtu` for unstructured meshes using VTU. That requires no libraries and is loadable in Paraview. I haven't tested it with cell-centered fields yet, but `cgns:out.cgns` has scalable parallel IO. (Back from SIAM CSE, but need some catching up.)

bishtgautam:
> We can do this, but it will force us to require NetCDF support in every installation in order to read the Exodus file. If that's okay, I'll add these files to the repo.

Yes, let's add those files to our repo and add a test for it. Additionally, please update the PETSc container if PETSc wasn't installed with exodus support.

bishtgautam:
After a discussion with @jeff-cohere, we decided to go with @jeff-cohere's suggestion of

- `RDyBoundary` - which will consist of a collection of edges
- `RDyRegion` - which will consist of a collection of cells

jeff-cohere:
Okay, I think this is now ready to go. Our Docker container already has Exodus II in it, and I've added tests for the file-based initial conditions.

bishtgautam (approved):

PR #33, opened 2023-04-25
Adds critical flow BC

bishtgautam:
Adds critical flow BC.

### DIFF
--- a/src/physics_swe.c
+++ b/src/physics_swe.c
@@ -136,22 +136,29 @@ static PetscErrorCode GetVelocityFromMomentum(PetscReal tiny_h, RiemannDataSWE *
   PetscFunctionReturn(0);
 }
 
-// Computes flux based on Roe solver
-// N    - Size of the array
-// hl   - Height left of the edge
-// hr   - Height right of the edge
-// ul   - Velocity in x-dir left of the edge
-// ur   - Velocity in x-dir right of the edge
-// vl   - Velocity in y-dir left of the edge
-// vr   - Velocity in y-dir right of the edge
-// sn   - sine of the angle between edge and y-axis
-// cn   - cosine of the angle between edge and y-axis
-// fij  - flux
-// amax - maximum courant number
-static PetscErrorCode ComputeRoeFlux(PetscInt N, const PetscReal hl[N], const PetscReal hr[N], const PetscReal ul[N], const PetscReal ur[N],
-                                     const PetscReal vl[N], const PetscReal vr[N], const PetscReal sn[N], const PetscReal cn[N], PetscReal fij[N][3],
-                                     PetscReal amax[N]) {
+/// Computes flux based on Roe solver
+/// @param [in] N Size of the array
+/// @param [in] *datal A RiemannDataSWE for values left of the edges
+/// @param [in] *datar A RiemannDataSWE for values right of the edges
+/// @param [in] sn Sine of the angle between edge and y-axis
+/// @param [in] cn Cosine of the angle between edge and y-axis
+/// @param [out] fij Flux through the edges
+/// @param [out] amax Maximum courant number
+/// @return 0 on success, or a non-zero error code on failure
+static PetscErrorCode ComputeRoeFlux(PetscInt N, RiemannDataSWE *datal, RiemannDataSWE *datar, const PetscReal sn[N], const PetscReal cn[N],
+                                     PetscReal fij[N][3], PetscReal amax[N]) {
   PetscFunctionBeginUser;
+
+  PetscReal *hl = datal->h;
+  PetscReal *ul = datal->u;
+  PetscReal *vl = datal->v;
+
+  PetscReal *hr = datar->h;
+  PetscReal *ur = datar->u;
+  PetscReal *vr = datar->v;
+
+  PetscAssert(datal->N == datar->N, PETSC_COMM_WORLD, PETSC_ERR_ARG_SIZ, "Size of data left and right of edges is not the same!");
+  PetscAssert(N == datal->N, PETSC_COMM_WORLD, PETSC_ERR_ARG_SIZ, "Size of data left/right of edges is not the same as sn/cn");
 
   for (PetscInt n = 0; n < N; n++) {
     // Compute Roe averages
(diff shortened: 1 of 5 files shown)


src/physics_swe.c
```diff
@@ -296,58 +296,51 @@ static PetscErrorCode RHSFunctionForInternalEdges(RDy rdy, Vec F, CourantNumberD
...
-    PetscInt iedge = boundary->edge_ids[e];
-    PetscInt icell = edges->cell_ids[2 * iedge];
+  PetscFunctionReturn(0);
+}
 
-    cn_vec_bnd[e] = edges->cn[iedge];
-    sn_vec_bnd[e] = edges->sn[iedge];
+// After the right values (hr/ur/vr) have been computed based on the different type of BCs,
+// compute the fluxes and add contribution in the F vector.
+static PetscErrorCode ComputeBC(RDy rdy, RDyBoundary *boundary, PetscReal tiny_h, CourantNumberDiagnostics *courant_num_diags, PetscInt N,
```
bishtgautam, line 328: @jeff-cohere The input argument list is absurdly too long and ideas on how to make it more readable are welcome. Maybe we can make a struct that contains the three variables`h`, `u` and `v`.

jeff-cohere: We could define a boundary condition struct that holds all this information, and initialize it with an appropriate function. For example, if we shoved all of the arrays of length `N` into a `SWEBCData` struct (for example):

```
typedef struct {
  // boundary associated with this data
  RDyBoundary *boundary;
  // number of boundary edges
  PetscInt N;
  // boundary edge data
  PetscReal *hl, *hr, *ul, *ur, *vl, *vr, *sn, *cn;
} BCDataSWE;

static PetscErrorCode BCDataSWECreate(RDyBoundary *boundary, BCDataSWE* bc_data) {
  PetscFunctionBegin;
  bc_data.boundary = boundary;
  PetscInt N = boundary->num_edges;
  bc_data.N = N;
  RDyAlloc(PetscReal, N, &boundary->hl);
  // etc
}

static PetscErrorCode BCDataSWEDestroy(BCDataSWE bc_data) {
  PetscFunctionBegin;
  RDyFree(bc_data.hl);
  // etc
}
```

Then `ComputeBC` might look like this:

```
static PetscErrorCode ComputeBC(RDy rdy, BCDataSWE* bc_data, PetscReal tiny_h, CourantNumberDiagnostics *courant_num_diags, PetscReal *X, PetscReal *F) {
  // etc
}
```

and `PerformPreComputationForBC` (or `PrecomputeBCData`, perhaps) would look similar.

Does this sound reasonable?

jeff-cohere: Of course, it might be good to chat with Jed and Matt sometime soon about their libceed QFunctions and their vectorization, in case we need to do anything special for our GPU-friendly strategy.

src/read_config_file.c
```diff
@@ -699,8 +699,8 @@ static PetscErrorCode ParseFlowConditions(yaml_event_t *event, YamlParserState *
       } else {
         if (!strcmp(state->parameter, "type")) {
           PetscInt selection;
-          SelectItem(value, 3, (const char *[3]){"dirichlet", "neumann", "reflecting"},
-                     (PetscInt[3]){CONDITION_DIRICHLET, CONDITION_NEUMANN, CONDITION_REFLECTING}, &selection);
+          SelectItem(value, 4, (const char *[4]){"dirichlet", "neumann", "reflecting", "critical-outflow"},
```
bishtgautam, line 702: @jeff-cohere  I used `critical-outflow` instead of using `critical_outflow`, which I think is the correct approach to avoid confusing yaml, but I want to double-check with you if it is the correct thing to do.

src/physics_swe.c
```diff
@@ -222,25 +267,27 @@ static PetscErrorCode RHSFunctionForInternalEdges(RDy rdy, Vec F, CourantNumberD
   PetscCheck(ndof == 3, rdy->comm, PETSC_ERR_USER, "Number of dof in local vector must be 3!");
 
   PetscInt  num = mesh->num_internal_edges;
-  PetscReal hl_vec_int[num], hul_vec_int[num], hvl_vec_int[num], ul_vec_int[num], vl_vec_int[num];
-  PetscReal hr_vec_int[num], hur_vec_int[num], hvr_vec_int[num], ur_vec_int[num], vr_vec_int[num];
   PetscReal sn_vec_int[num], cn_vec_int[num];
   PetscReal flux_vec_int[num][3], amax_vec_int[num];
 
+  RiemannDataSWE datal, datar;
+  PetscCall(RiemannDataSWECreate(num, &datal));
```
jeff-cohere, line 274: Here we are allocating and deleting memory every time we want to compute fluxes. If this causes too much "churn" (asking the OS for memory and then handing it back a lot), there are things we can do to alleviate it. But I think it's fine for now--just making a note.

src/physics_swe.c
```diff
@@ -293,61 +339,56 @@ static PetscErrorCode RHSFunctionForInternalEdges(RDy rdy, Vec F, CourantNumberD
...
 }
 
-// applies a reflecting boundary condition on the given boundary, computing
-// fluxes F for the solution vector components X
-static PetscErrorCode ApplyReflectingBC(RDy rdy, RDyBoundary *boundary, PetscReal tiny_h, CourantNumberDiagnostics *courant_num_diags, PetscReal *X,
-                                        PetscReal *F) {
+// Before computing BC fluxes, perform common precomputation irrespective of BC type that include:
+// (i) extracting h/hu/hv from the solution vector X, and
+// (ii) compute velocities (u/v) from momentum (hu/hv).
+static PetscErrorCode PerformPreComputationForBC(RDy rdy, RDyBoundary *boundary, PetscReal tiny_h, PetscInt N, RiemannDataSWE *data, PetscReal cn[N],
```
jeff-cohere, line 351: What do you think about renaming this to `PrecomputeBC`? Again, not a big deal. :-)

bishtgautam: done.

src/physics_swe.c
bishtgautam: Good point. I have added comments about it.

jeff-cohere (approved):
This looks good to me. I'll think about how to minimize OS churn with allocation/freeing of Riemann solver data, since it seems like we're doing it in a lot of places. But I don't think it's urgent at the moment. And some of the solutions to this problem (e.g. an allocator that holds on to a certain pool of memory) would allow us to keep this code as is.

Are you concerned about the level of test coverage? If so, we can probably add some more unit tests.

bishtgautam:
To improve the code coverage since I initially opened the PR, I have now [updated the ex2b.yaml](https://github.com/RDycore/RDycore/pull/33/files#diff-739c44ee05b05444bd5d2a30ce93216bb4fa14409fbe193113d7a512e98b37ed) to include a critical flow BC. So, I'm not concerned with current code coverage.

PR #36, opened 2023-05-11
Implemented XDMF output

### LINKED ISSUE #27: Add XDMF output capability
Now that we've got a simulator that can run alongside our toy problem demos, it's a good time to add some features for inspecting output.

jeff-cohere:
This PR adds an XDMF output capability to RDycore, along with a stubbed-out CGNS option, which could be added after some related code gets merged to PETSc.

Currently, all data is written to a single "heavy data" HDF5 file, and .xmf XML "light data" files are generated for each desired output step, which allows us to view results before the end of a simulation.

Also included is a change to the determination of boundary edges [suggested by Jed](https://github.com/RDycore/RDycore/issues/27#issuecomment-1529254151) that excludes boundary vertices.

Open to suggestions as to how we test this capability. Maybe we need to focus on a verification workflow next?

## Caveats

* XDMF output has not been tested on meshes with multiple topologies (e.g. triangles + quads). I'll do this in the next PR
* This PR introduces a hack for `FindPETSc.cmake` that will get us what we need until PETSc has an official CMake solution.

Closes #27


bishtgautam (approved):
This is good to go.

PR #46, opened 2023-05-22
Adds support for CGNS output and I/O-related enhancements

### LINKED ISSUE #35: Add enhancements to xdf output following PFLOTRAN's approach
1. PFLOTRAN has a nice feature regarding `.xdf` output that controls how many time slices 
are written in a `.h5` and this feature is governed by the following in PFLOTRAN inputdeck:

```
FORMAT HDF5 MULTIPLE_FILES TIMES_PER_FILE 10
```

In the inputdeck mentioned above, PFLOTRAN will write 10 time slices in a `.h5` file before 
opening another .h5.

2. PFLOTRAN has the following cards in inputdecks that control periodic outputs

```
PERIODIC TIMESTEP <int>
Generates output every <int> number of timesteps.

PERIODIC TIME <int or float> <string>
Generates output at every <int or float> units of time, where <string> defines the units of time.
```

PFLOTRAN's documentation on the output card is available [here](https://www.pflotran.org/documentation/user_guide/cards/subsurface/output_card.html?highlight=output
).

### LINKED ISSUE #40: Support CGNS output
Now that @jedbrown 's support for HDF5-based CGNS files has made it into PETSc proper, we can easily add support for it to RDycore.

jeff-cohere:
This PR adds support for CGNS output using PETSc's CGNS viewer. It also adds support for output "batching" via a new output parameter called `batch_size`, e.g.

```
output:
    format: cgns
    frequency: 100
    batch_size: 10
```

All HDF5-based data files (XDMF "heavy" data and CGNS data) can hold data for a number of steps equal to its batch size. Periodic output is supported via the `frequency` parameter. The `frequency` value of 100, above, means that output is dumped every 100 steps. We could use the term `interval` here instead if `frequency` is too confusing (being the reciprocal of the quantity of interest).

Issue #35 mentions the desire for periodic output based on solution time in addition to number of steps. I've seen this done and implemented it in other codes, but it seems like visualization software can interpolate between steps to emulate this behavior, and also we'd need to figure out how to dump output without affecting the timestep that is optimal for the numerical solution of the problem, which seems like it could be complicated. I welcome everyone's thoughts on this.

I've verified that I can view CGNS output from serial and parallel runs in Paraview. Visit seems to have some issues with this output. I've relied on PETSc's options database exclusively for CGNS output, but I don't know if I've missed something.

**EDIT: I've also added a simple config file for CodeCov that sets our coverage acceptance threshold at 90% and disables "patch coverage". I think this is a decent figure for us at the moment, but we can revise it if we want.**

Closes #35 
Closes #40


jedbrown:
Hmm, I've opened some CGNS files in VisIt, but mostly use Paraview myself. VisIt choked on high order (I think cubic and up), but this is different. I can explore if you'd like. (I don't know how heavily this project uses VisIt. It's a possible upstream VisIt needs a bug fix.)

jeff-cohere:
Thanks, @jedbrown . If you have an angle in mind to explore this using Visit, I'd certainly appreciate your thoughts. I'm more familiar with Visit but I don't mind using Paraview--it's gotten much more stable than it used to be.

bishtgautam:
@jeff-cohere For CNGS output, I noticed that the names of the variables are `height`, `x_momentum`, and `y_momentum`.
cgns-visit

I wanted to change the variables and I made the following code modifications, but nothing changed when I loaded the files in VisIt and Paraview.
```
diff --git a/src/cgns_output.c b/src/cgns_output.c
index e2fb70e..217900c 100644
--- a/src/cgns_output.c
+++ b/src/cgns_output.c
@@ -41,9 +41,9 @@ PetscCall(DMPlexGlobalToNaturalEnd(rdy->dm, rdy->X, natural));
 // FIXME: This setup is specific to the shallow water equations. We can
 // FIXME: generalize it later.
 const char *comp_names[3] = {
-    "Water_Height",
-    "X_Momentum",
-    "Y_Momentum",
+    "WaterHeight",
+    "XMomentum",
+    "YMomentum",
 };
 Vec      comp;  // single-component natural vector
 PetscInt n, N, bs;
diff --git a/src/write_xdmf_output.c b/src/write_xdmf_output.c
index 0cff368..a700524 100644
--- a/src/write_xdmf_output.c
+++ b/src/write_xdmf_output.c
@@ -52,9 +52,9 @@ static PetscErrorCode WriteXDMFHDF5Data(RDy rdy, PetscInt step, PetscReal time)
   // FIXME: This setup is specific to the shallow water equations. We can
   // FIXME: generalize it later.
   const char *comp_names[3] = {
-      "Water_Height",
-      "X_Momentum",
-      "Y_Momentum",
+      "WaterHeight",
+      "XMomentum",
+      "YMomentum",
   };
   Vec      comp;  // single-component natural vector
   PetscInt n, N, bs;
@@ -231,7 +231,7 @@ static PetscErrorCode WriteXDMFXMFData(RDy rdy, PetscInt step, PetscReal time) {
     // (none so far!)

     // write cell field metadata
-    const char *cell_field_names[3] = {"Water_Height", "X_Momentum", "Y_Momentum"};
+    const char *cell_field_names[3] = {"WaterHeight", "XMomentum", "YMomentum"};
     for (int f = 0; f < 3; ++f) {
       PetscCall(PetscFPrintf(rdy->comm, fp,
                              "      <Attribute Name=\"%s\" AttributeType=\"Scalar\" Center=\"Cell\">\n"
```

Where are the names of these variables in the CNGS being defined?

bishtgautam (commented):
let's update one of the example to output cgns

jedbrown:
We should avoid this duplication, but these are the lines to visit
```
src/rdysetup.c
150:  char         field_names[3][20] = {"height", "x momentum", "y momentum"};
156:    PetscCall(PetscSectionSetFieldName(sec, f, &field_names[f][0]));
```
If you name them like `MomentumX`, `MomentumY`, then they'll appear automatically as vectors in Paraview.

Since these all live in the same space (cell-centered constants), I'd encourage making them one field with three components instead of three fields (this simplifies the integer data structures).

bishtgautam:
>Since these all live in the same space (cell-centered constants), I'd encourage making them one field with three components instead of three fields (this simplifies the integer data structures).

Would the following code change achieve what you are advocating for?

```
  // create a section with (h, hu, hv) as degrees of freedom
  PetscInt     n_field            = 1;
  PetscInt     n_field_dof[n_field]     = {3};
  char         field_names[n_field][20] = {"SWEPrognosticVars"};
  PetscSection sec;
  PetscCall(PetscSectionCreate(rdy->comm, &sec));
  PetscCall(PetscSectionSetNumFields(sec, n_field));
  PetscInt n_field_dof_tot = 0;
  for (PetscInt f = 0; f < n_field; ++f) {
    PetscCall(PetscSectionSetFieldName(sec, f, &field_names[f][0]));
    PetscCall(PetscSectionSetFieldComponents(sec, f, n_field_dof[f]));
    n_field_dof_tot += n_field_dof[f];
  }
```

jedbrown:
Yeah. You don't need to set a field name, but would use `PetscSectionSetComponentName(sec, f, c, comp_name[c])` to set the three component names "Thickness", "MomentumX", and "MomentumY". BTW, do you intend for the first component to be thickness (conservative) or height (not conservative on a non-flat bed).

jeff-cohere:
I'll also make the changes to the component names once we clarify whether we're using height or thickness. I wasn't aware of their nonconservative/conservative relationship.

bishtgautam:
Currently, we are using water height.

jeff-cohere:
Okay, I think this PR is ready now. Thanks for your patience, everyone.

bishtgautam:
While writing XDMF, we get the following information on the command line:

```
DETAIL: Step 0: writing XDMF HDF5 output at t = 0. hr to output/ex2b-
DETAIL: Step 0: writing XDMF XMF output at t = 0. hr to output/ex2b-0000.xmf
DETAIL: Step 100: writing XDMF HDF5 output at t = 0.0005 hr to output/ex2b-
DETAIL: Step 100: writing XDMF XMF output at t = 0.0005 hr to output/ex2b-0100.xmf
```

For CGNS, we are adding `-ts_monitor_solution`. Can we also print info about the cgns output file on command line?

jeff-cohere:
That's a good point. I can do this by adding a function that just prints out this message, since the CGNS method is handled exclusively by command line arguments.

bishtgautam (approved):
@jeff-cohere Thanks for the multiple iterations. This PR is ready to be merged.

PR #50, opened 2023-06-01
Restructured the YAML parser to use schemas

jeff-cohere:
Since we've been adding more features to RDycore and elaborating on our input format, and because we plan to keep adding such features, I decided to make the YAML parser a bit easier to work with.

This PR pulls in a relatively new library called `libcyaml` that allows us to define schemas for the sections in our output specification. In particular, the parser now checks an input file against this schema and reports any problems in a traceback like this:

```
Load: Unexpected key: 1
Load: Backtrace:
  in mapping (line: 19, column: 3)
  in mapping field 'initial_conditions' (line: 19, column: 3)
Error encountered in ParseYaml (/home/jeff/projects/pnnl/RDycore/src/read_config_file.c, line 430): Error parsing config file: Invalid key (83)
```

This should make it much easier for us to verify that we're running simulations the way we think we are. As a bonus, the parser code is now much shorter and hopefully easier to understand.

Additionally, the `make format-c` target has been extended to the `include` folder to make sure our header files are properly formatted.

## Updates to Input Spec

Because we've moved to a schema-based input model, it made sense to introduce some changes to a few input sections (which I've updated in our test inputs):

1. The `initial_conditions` and `sources` sections have been modified to include a `domain` subsection in which you can optionally specify a file/format for reading initial conditions and sources for the entire domain (see [here](https://github.com/RDycore/RDycore/blob/jeff-cohere/cyaml-parser/driver/tests/swe_roe/ex2b_ic_file.yaml)). There's also a `regions` subsection in which conditions are listed as a YAML sequence with the region's `id` and the usual physics-related condition names (see [here](https://github.com/RDycore/RDycore/blob/jeff-cohere/cyaml-parser/driver/tests/swe_roe/ex2b.yaml)).
2. The `boundary_conditions` section uses the same sequence-based approach as the aforementioned sections for listing named boundary conditions.
3. The `flow_conditions`, `sediment_conditions`, and `salinity_conditions` sections also use YAML sequences listing their `name`s and all relevant parameters.
4. Minor changes have been made to other sections. Take a look at the linked input files to see what you think.

If we like this format, it should be much easier to modify it in the future. I'll update our input spec in Confluence if this PR is approved.


driver/tests/swe_roe/ex2b.yaml
```diff
@@ -41,22 +42,21 @@ initial_conditions:
 # the BC for "boundary". If no BC is specified for a boundary, the reflecting
 # wall BC is assumed.
 boundary_conditions:
-  2: # top_wall
+  - id: 2 # top_wall
```
bishtgautam, line 45: addition of `id` make it intuitive. 👍

driver/tests/swe_roe/ex2b.yaml
```diff
@@ -41,22 +42,21 @@ initial_conditions:
...
     flow: reflecting_bc
-  3: # bottom_wall
+  - id: 3 # bottom_wall
     flow: outflow_bc
 
 flow_conditions:
-  dam_top_ic:
+  - name: dam_top_ic
```
bishtgautam, line 51: similarly adding `name` is helpful.

driver/tests/swe_roe/ex2b_ic_file.yaml
```diff
@@ -26,6 +26,6 @@ grid:
   file: DamBreak_grid5x10.exo
 
 initial_conditions:
-  format: binary
-  file: DamBreak_grid5x10_wetdownstream.IC
-
+  domain:
```
bishtgautam, line 29: The use `domain` is great as it tells this IC is applied on the entire domain.

bishtgautam (approved):
Thanks @jeff-cohere, this is great.

PR #54, opened 2023-06-07
Add libCEED to evaluate RHS

jedbrown:
This PR adds GPU support via libCEED and PETSc.

Run a 135k cell dam break using libCEED on CPU
```console
$ ../../rdycore ex2b.yaml -log_view -log_view_gpu_time -ts_monitor_solution cgns:s.cgns -dm_refine 5 -ts_dt .005 -ts_monitor -ts_max_time 1 -ts_type ssp -ceed /cpu/self
```

and on GPU
```console
$ ../../rdycore ex2b.yaml -log_view -log_view_gpu_time -ts_monitor_solution cgns:s.cgns -dm_refine 5 -ts_dt .005 -ts_monitor -ts_max_time 1 -ts_type ssp -ceed /gpu/cuda -dm_vec_type cuda
```
This video was produced by a GPU run.

[dam-break.webm](https://github.com/RDycore/RDycore/assets/3303/e6788d27-a907-481d-bdf0-60a10da5c4e9)

## Caveats
* Do we want libCEED to be an optional dependency? If so, we need a little preprocessor or linking shenanigans.
* I have not yet implemented boundary conditions on device so you'll see host<->device transfers in the profile. Disabling boundary conditions will result in no such traffic (but you don't get correct BCs). I'm not really a fan of packing ghost cells if we can get away from it; I'd rather just initialize the exterior state within the qfunction (this is what we do in ceed-fluids). We can add this easily once we agree on strategy. **EDIT: Boundary conditions are implemented.**
* Source terms are not yet implemented in libCEED.
* I am not returning CFL numbers here. Doing so requires a reduction, which is somewhat expensive for the device and parallel communication (will be visible in scaling studies). And I'm not sure it's actually needed at each stage in the time integrator. Perhaps once per step (`-ts_type ssp`, which I recommend, does several stages per step).


src/swe-flux.h
```diff
@@ -0,0 +1,111 @@
...
+#define swe_flux_h
+
+#include <ceed/types.h>
+
+struct SWEState_ {
+  CeedScalar h, hu, hv;
+};
+typedef struct SWEState_ SWEState;
+
+CEED_QFUNCTION_HELPER void SWERiemannFlux_Roe(SWEState qL, SWEState qR, CeedScalar sn, CeedScalar cn, CeedScalar flux[], CeedScalar *amax) {
```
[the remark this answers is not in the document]
jeff-cohere: That's great. Personally, I'd be happy with just using CEED. @bishtgautam , any opinion on this?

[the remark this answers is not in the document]
knepley: I think, at least in the short term, we should keep the code, if only to show that we get significant speedup.

knepley:
@arcowie Can you review this as well, and make sure we can reproduce locally?

src/physics_swe.c
```diff
@@ -3,6 +3,8 @@
 #include <private/rdymemoryimpl.h>
 #include <stddef.h>  // for offsetof
 
+#include "swe-flux.h"
```
jeff-cohere, line 6: Picking a nit: would you mind if we called this `swe_flux.h`? We have some files already that use underscores, and as a victim of the trauma created by C++'s "frivolous decision creation syndrome", I'm sensitive to having multiple ways of naming and referring to things. :-)

bishtgautam:
It is okay for us to require PETSc to be compiled with libCEED. 

@jeff-cohere Can you please update the docker image to also download libCEED?

jeff-cohere:
> It is okay for us to require PETSc to be compiled with libCEED.
> 
> @jeff-cohere Can you please update the docker image to also download libCEED?

Sure, and I agree--this project is E3SM-focused, so we have a lot of control over what we require of users.

jeff-cohere (approved):
This looks great to me--I left a comment on a filename. I'll make sure we have CEED available in our CI environment.

jeff-cohere:
Hmm. Apparently, I added `--download-libceed` to the PETSc build in our Docker image back in December, so it should be in there already. @jedbrown , any thoughts on what we might need to do to point CMake at it?

jedbrown:
In that case `PKG_CONFIG_PATH=$PETSC_DIR/$PETSC_ARCH/lib/pkgconfig` should be sufficient.

jeff-cohere:
Okay, I think I've patched in PETSc's install of libceed (commit to follow). However, `swe-flux.h` asks for a file that doesn't seem to be included in the installation:

```
#include <ceed/types.h>
```

Here's what's in the `ceed/` header directory:

```
jeff@trippy:~/projects/pnnl/petsc/debug/include/ceed$ ls
backend.h  ceed-f32.h  ceed-f64.h  ceed.h  cuda.h  fortran.h  hash.h  hip.h  khash.h
```

Is this expected?

jedbrown:
Okay, the issue is that the libCEED version downloaded by PETSc has gotten stale. This will get y'all running and I'll make a PR to PETSc to update it.
```diff
diff --git i/config/BuildSystem/config/packages/libceed.py w/config/BuildSystem/config/packages/libceed.py
index 7ee02173d69..beb37337bb6 100644
--- i/config/BuildSystem/config/packages/libceed.py
+++ w/config/BuildSystem/config/packages/libceed.py
@@ -3,7 +3,7 @@ import config.package
 class Configure(config.package.Package):
   def __init__(self, framework):
     config.package.Package.__init__(self, framework)
-    self.gitcommit              = 'd086b42987308d3f7585a8cbf1070065d766cfaa' # v0.10.0 on Mar 21, 2022
+    self.gitcommit              = '7c36fa1d7d7dd3298f8241b8782ca5c7e472d0eb' # v0.11.0-307-g7c36fa1d7 on June 5, 2023
     self.download               = ['git://https://github.com/CEED/libceed.git']
     self.functions              = ['CeedRegister']
     self.includes               = ['ceed.h']
```

arcowie:
That worked and my movie looks like @jedbrown 's movie.  Results with ceed on the CPU
Summary of Stages:   ----- Time ------  ----- Flop ------  --- Messages ---  -- Message Lengths --  -- Reductions --
                        Avg     %Total     Avg     %Total    Count   %Total     Avg         %Total    Count   %Total
 0:      Main Stage: 4.3032e+01 100.0%  3.5144e+08 100.0%  0.000e+00   0.0%  0.000e+00        0.0%  0.000e+00   0.0%

------------------------------------------------------------------------------------------------------------------------

When I try to run it using cuda it fails.  I will install the drivers and then reconfigure PETSc --with-cuda and compare the performance to see if I get the same speed up.

CMakeLists.txt
```diff
@@ -20,9 +20,11 @@ project(rdycore)
 set(PETSC_DIR $ENV{PETSC_DIR})
 set(PETSC_ARCH $ENV{PETSC_ARCH})
 find_package(PETSc COMPONENTS C)
+list(APPEND CMAKE_PREFIX_PATH ${PETSC_DIR}/${PETSC_ARCH}/lib/pkgconfig)
 
 include(FindPkgConfig)
 pkg_check_modules(CEED REQUIRED IMPORTED_TARGET ceed)
+include_directories(${CEED_INCLUDE_DIRS})
```
jedbrown, line 27: This line already exists down below next to this:
```
include_directories(${PETSC_INCLUDES})
```
Can we delete one or the other?

jeff-cohere: Sure, I'll make sure we don't need the `CEED_INCLUDE_DIRS`. I'm having some issues getting the CI environment to work with this PR, but I'll prune this when I've figured it out.

src/physics_swe.c
jedbrown: Sure, are you still hacking on the branch or should I just push?

jeff-cohere: Go ahead and push. Thanks!

jeff-cohere: Okay, now I'm definitely finished!

jeff-cohere:
Sorry for all the noise, folks. This PR is all set within the CI environment and ready for your review.

jedbrown:
I added CEED boundary conditions. We currently have some host<->device transfers as noted in this issue (Cc: @knepley).
https://gitlab.com/petsc/petsc/-/issues/1395

Options to remove these transfers:
1. Fix that issue to so `INSERT_VALUES` can be all-GPU like `ADD_VALUES` is now.
2. Compute fluxes only on "owned" edges and use `ADD_VALUES`
3. Split the volume operator into interior (which contributes 2-sided flux) and interface (which applies flux only to the owned cell, not the ghost cell), deposit result directly in global vector. This would also remove what is effectively a vector copy, and would allow overlap of communication and computation at the expense of an extra kernel launch. On balance, it's probably a worthwhile optimization, but I'd rather it be separate from this PR.

bishtgautam (commented):
How about the following?
- Renaming `swe_flux.h` to be `swe_flux_ceed.h` to indicate it has CEED code, and
- Pulling out corresponding functions from `physics_swe.c` and putting them in `swe_flux_petsc.h`

Later on, we can also name the variables in `swe_flux_ceed.h` and `swe_flux_petsc.h` to be as same as possible.

src/swe-flux.h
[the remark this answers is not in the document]
bishtgautam: I think we should keep the non-CEED code for the time being till the time we build more expertise on the team to directly code in CEED.

CMakeLists.txt
```diff
@@ -20,8 +20,7 @@ project(rdycore)
 set(PETSC_DIR $ENV{PETSC_DIR})
 set(PETSC_ARCH $ENV{PETSC_ARCH})
 find_package(PETSc COMPONENTS C)
-set(ENV{PKG_CONFIG_PATH} ${PETSC_DIR}/${PETSC_ARCH}/lib/pkgconfig)
-list(APPEND CMAKE_PREFIX_PATH $ENV{PKG_CONFIG_PATH})
+set(ENV{PKG_CONFIG_PATH} "$ENV{PKG_CONFIG_PATH}:${PETSC_DIR}/${PETSC_ARCH}/lib/pkgconfig")
```
jedbrown: @jeff-cohere I hope you're okay with this. The other prevents using libCEED that isn't configured by PETSc.

jedbrown:
I implemented option 2. This should be ready for parallel testing on Perlmutter or Crusher.

jedbrown:
0.5 microseconds/RHS evaluation with 2M cells. This profile is showing the vector work taking more time than RHS evaluation, which I think must be a bit skewed. In any case, I don't expect that to persist in parallel runs. We should make an event for the local-only part (the `CeedOperatorApply`).
```console
$ ./../rdycore ex2b.yaml -log_view -log_view_gpu_time -ts_monitor_solution cgns:s.cgns -ts_monitor_solution_interval 10 -dm_re
fine 7 -ts_dt .00125 -ts_monitor -ts_max_time 1 -ts_type ssp -ceed /gpu/cuda -dm_vec_type cuda -use_gpu_aware_mpi 0
[...]
CeedElemRestriction from (2162688, 3) to 1439488 elements with 1 nodes each and component stride 1
[...]
VecAXPY             3200 1.0 2.0373e+00 1.0 1.38e+10 1.0 0.0e+00 0.0e+00 0.0e+00 12 62  0  0  0  12 62  0  0  0  6794   63554      0 0.00e+00    0 0.00e+00  100
VecAXPBYCZ           800 1.0 6.2167e-01 1.0 8.65e+09 1.0 0.0e+00 0.0e+00 0.0e+00  4 38  0  0  0   4 38  0  0  0 13915   58619      0 0.00e+00    0 0.00e+00  100
VecCUDACopyTo          2 1.0 5.5851e-05 1.0 0.00e+00 0.0 0.0e+00 0.0e+00 0.0e+00  0  0  0  0  0   0  0  0  0  0     0       0      2 3.46e+01    0 0.00e+00  0
VecCUDACopyFrom       83 1.0 1.6341e-01 1.0 0.00e+00 0.0 0.0e+00 0.0e+00 0.0e+00  1  0  0  0  0   1  0  0  0  0     0       0      0 0.00e+00   83 1.44e+03  0
cuBLAS Init            1 1.0 6.7908e-01 1.0 0.00e+00 0.0 0.0e+00 0.0e+00 0.0e+00  4  0  0  0  0   4  0  0  0  0     0       0      0 0.00e+00    0 0.00e+00  0
DCtxCreate             1 1.0 3.9620e-05 1.0 0.00e+00 0.0 0.0e+00 0.0e+00 0.0e+00  0  0  0  0  0   0  0  0  0  0     0       0      0 0.00e+00    0 0.00e+00  0
DCtxSetUp              1 1.0 2.1570e-05 1.0 0.00e+00 0.0 0.0e+00 0.0e+00 0.0e+00  0  0  0  0  0   0  0  0  0  0     0       0      0 0.00e+00    0 0.00e+00  0
DCtxSetDevice          1 1.0 7.2941e-05 1.0 0.00e+00 0.0 0.0e+00 0.0e+00 0.0e+00  0  0  0  0  0   0  0  0  0  0     0       0      0 0.00e+00    0 0.00e+00  0
DCtxSync           10490 1.0 1.1034e-01 1.0 0.00e+00 0.0 0.0e+00 0.0e+00 0.0e+00  1  0  0  0  0   1  0  0  0  0     0       0      0 0.00e+00    0 0.00e+00  0
TSStep               800 1.0 5.7413e+00 1.0 2.25e+10 1.0 0.0e+00 0.0e+00 0.0e+00 33 100  0  0  0  33 100  0  0  0  3918   55023      0 0.00e+00    0 0.00e+00  100
TSFunctionEval      4000 1.0 2.3445e+00 1.0 0.00e+00 0.0 0.0e+00 0.0e+00 0.0e+00 13  0  0  0  0  13  0  0  0  0     0       0      0 0.00e+00    0 0.00e+00  0
```

bishtgautam:
@jedbrown Can you rebase this branch off the latest `main`?

bishtgautam:
Or, can I push the rebased branch?

jedbrown:
I implemented the rename you requested and will push with the rebase in a minute.

bishtgautam:
I'm using PETSc hash `99e66fd4349` and it doesn't include `-ts_monitor_solution cgns`. What is the PETSc hash that you are using?

jedbrown:
Just need to configure with `--download-cgns --download-hdf5`.

jedbrown:
So I need to turn off the verbose printing of libCEED stuff. Do we have a pattern for making it optional or should I just comment it out (a couple lines) so others can remember how to view it?

I think we should print some diagnostics including the Vec type and the CEED resource being used in the run.

jedbrown:
I pushed a commit to `if (0)` the printing -- we can resolve verbose printing in a later PR. I think this PR is ready to merge when @bishtgautam approves.

bishtgautam (approved):

PR #60, opened 2023-06-21
Adds four-mounds problem

bishtgautam:
Adds a four-mound problem, which can be used for developing the libCEED version of the `AddSourceTerm()` function.

**Initial condition**
four_mounds0000

**Final solution**
four_mounds0049


bishtgautam:
The above results were produced using:

```
cd build/driver/tests/swe_roe

# Edited `final_time: 0.5` to be `final_time: 10.0`

../../rdycore four_mounds_60x24.yaml
```

jedbrown (commented):
Does this compute the wrong result when using `-ceed /cpu/self` because that lacks a source operator? Have we decided whether to do the splitting of source and Riemann flux as separate operators or to put them in the same one?

driver/tests/swe_roe/CMakeLists.txt
```diff
@@ -30,3 +30,16 @@ foreach(np 1 2) # test on 1, 2 processes
...
+file (COPY
+      ${CMAKE_CURRENT_SOURCE_DIR}/four_mounds_60x24.yaml
+      ${MESH_DIR}/four_mounds_60x24.exo
+      ${IC_DIR}/four_mounds_60x24.ic
+      ${MATERIALS_DIR}/manning_four_mound_60x24.bin
+      DESTINATION ${CMAKE_CURRENT_BINARY_DIR})
+foreach(np 1 2)
+  add_test(swe_roe_four_mounds_c_np_${np} ${PETSC_MPIEXEC} ${PETSC_MPIEXEC_FLAGS} -n ${np} ${rdycore_driver} four_mounds_60x24.yaml)
+  add_test(swe_roe_four_mounds_f90_np_${np} ${PETSC_MPIEXEC} ${PETSC_MPIEXEC_FLAGS} -n ${np} ${rdycore_f90_driver} four_mounds_60x24.yaml)
+  add_test(swe_roe_four_mounds_f90_np_${np}_ceed ${PETSC_MPIEXEC} ${PETSC_MPIEXEC_FLAGS} -n ${np} ${rdycore_f90_driver} four_mounds_60x24.yaml -ceed /cpu/self)
```
bishtgautam, line 44: I should not have added this and I will remove it.

bishtgautam:
> Does this compute the wrong result when using -ceed /cpu/self because that lacks a source operator? 

After the first Euler time step, there are a lot of `Inf` values within the domain.

>Have we decided whether to do the splitting of source and Riemann flux as separate operators or to put them in the same one?

I don't recall discussing this earlier. How would you combine the operators because they are looping over different mesh elements (edges vs cells)? If the operators are combined, would each Riemann flux operator, one would add the source/sink term for the left and right cell divided by the number of neighbors of the left and right cells, respectively?

jedbrown:
What is our approach to well-balancing? I haven't implemented this method from Randy LeVeque, but my understanding is that these methods usually store bathymetry as cell-centered and apply the source as part of the numerical flux. (There may be more nuance to a vertex-based bathymetry. I haven't researched recent work on this.)
https://doi.org/10.1007/s10915-010-9411-0

image

bishtgautam:
@jedbrown We don't have an approach for a well-balance scheme yet, but we need to do lit. review for it.

jedbrown:
Okay. It wouldn't be a lot of work to create a `CeedOperator` for the non-well-balanced scheme, but it's also hard to motivate writing that code if we're going to replace it with a face-based well-balancing method. If you're okay merging this PR without a ceed implementation, go ahead (it's a useful test) and we can work out balancing (and wetting/drying) in a follow-up.

bishtgautam:
The motivation for this PR is to add this example and the `CeedOperator` would be developed at a later stage.

jedbrown (approved):

PR #66, opened 2023-07-19
Adds a source-term to SWE.

bishtgautam:
- Adds a water source/sink term in the SWE. The Vec to hold the water source/sink
  is a 1-DOF Vec.
- Adds function/subroutine for an external model (e.g., E3SM) to specify the water source/sink term.
- The source/sink term is currently only added in the PETSc-version of the code.
- The two drivers are updated to call the new function/subroutine.

### DIFF
--- a/include/private/rdycoreimpl.h
+++ b/include/private/rdycoreimpl.h
@@ -122,7 +122,7 @@ struct _p_RDy {
   Vec R;
 
   // source-sink vector
-  Vec S;
+  Vec WaterSrc;
 
   //-------------------
   // Simulatіon output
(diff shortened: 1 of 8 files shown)


include/private/rdycoreimpl.h
```diff
@@ -121,6 +121,9 @@ struct _p_RDy {
   // residual vector
   Vec R;
 
+  // source-sink vector
+  Vec WaterSrc;
```
jeff-cohere, line 125: Could we call this `water_src` instead?

jeff-cohere (approved):
Looks great. I left one comment about the name of the source vector in the `RDy` struct.

PR #67, opened 2023-07-24
Implemented boundary flux output

### LINKED ISSUE #63: Diagnostic output: boundary fluxes
For the Houston Harvey flooding experiments, we would like to generate a tab-delimited time series file containing accumulated boundary fluxes for the components of the solution to the shallow water equations. The following columns would appear in such a file:

1. simulation time (in desired units)
2. (global) index of boundary edge
3. integer ID indicating the boundary to which the boundary edge belongs
4. integer ID indicating the type of boundary condition on the boundary edge
6. water mass flux
7. x-momentum flux
8. y-momentum flux
9. boundary edge normal x component
10. boundary edge normal y component

Boundary edges that are not explicitly assigned boundary conditions in the input YAML file do not appear in this file.

The file consists of rows containing the above information, with fluxes accumulated since the preceding row was printed. MPI rank 0 gathers all necessary information and appends a new row to the file at the desired dump interval as specified in the YAML file.

This means we'll need a parameter in the YAML file that defines the number of steps between boundary flux rows being appended to the file. We can do this by adding a subsection to the `output` section of the file--something like

```
output:
    ...
    time_series:
        boundary_fluxes: <number of steps between accumulated fluxes>
```

@bishtgautam

jeff-cohere:
I've got the boundary flux output capability working for non-CEED SWE fluxes. If we like what we see here, I can poke around with CEED and see whether/how I need to change how we accumulate boundary fluxes.

Also, I've changed `output.frequency` to `output.interval` to match the `coupling_interval` parameter.

Closes #63

### DIFF
--- a/src/time_series.c
+++ b/src/time_series.c
@@ -1,0 +1,95 @@
+#include <private/rdycoreimpl.h>
+#include <private/rdymemoryimpl.h>
+
+static PetscErrorCode InitBoundaryFluxes(RDy rdy) {
+  PetscFunctionBegin;
+
+  // count up boundary fluxes and allocate storage.
+  PetscInt num_boundary_fluxes = 0;
+  for (PetscInt b = 0; b < rdy->num_boundaries; ++b) {
+    RDyCondition bc = rdy->boundary_conditions[b];
+    if (!bc.auto_generated) {  // exclude auto-generated boundary conditions
+      RDyBoundary *boundary = &rdy->boundaries[b];
+      num_boundary_fluxes += boundary->num_edges;
+    }
+  }
+  PetscCall(RDyAlloc(PetscReal, 3 * num_boundary_fluxes, &rdy->time_series.boundary_fluxes));
+
+  PetscFunctionReturn(0);
+}
+
+// Initializes time series data storage.
+PetscErrorCode InitTimeSeries(RDy rdy) {
+  PetscFunctionBegin;
+  if (rdy->config.output.time_series.boundary_fluxes > 0) {
+    InitBoundaryFluxes(rdy);
+    PetscCall(TSMonitorSet(rdy->ts, WriteTimeSeries, rdy, NULL));
+  }
+  PetscFunctionReturn(0);
+}
+
+static PetscErrorCode WriteBoundaryFluxes(RDy rdy, PetscReal time) {
+  PetscFunctionBegin;
+  char dir[PETSC_MAX_PATH_LEN];
+  PetscCall(GetOutputDir(rdy, dir));
+  char path[PETSC_MAX_PATH_LEN];
+  snprintf(path, PETSC_MAX_PATH_LEN - 1, "%s/boundary_fluxes.dat", dir);
+  FILE *fp;
+  PetscCall(PetscFOpen(rdy->comm, path, "a", &fp));  // append to end of file
+  PetscInt n = 0;                                    // offset index for time series data
+  for (PetscInt b = 0; b < rdy->num_boundaries; ++b) {
+    RDyCondition bc = rdy->boundary_conditions[b];
+    if (!bc.auto_generated) {  // exclude auto-generated boundary conditions
+      RDyBoundary *boundary = &rdy->boundaries[b];
+      for (PetscInt e = 0; e < boundary->num_edges; ++e, ++n) {
+        PetscInt         edge_id        = boundary->edge_ids[e];
+        PetscInt         global_edge_id = rdy->mesh.edges.global_ids[edge_id];
+        RDyVector        edge_normal    = rdy->mesh.edges.normals[edge_id];
+        PetscInt         boundary_id    = rdy->boundary_ids[b];
+        RDyConditionType bc_type        = bc.flow->type;
+        PetscReal        water_mass     = rdy->time_series.boundary_fluxes[n].water_mass;
+        PetscReal        x_momentum     = rdy->time_series.boundary_fluxes[n].x_momentum;
+        PetscReal        y_momentum     = rdy->time_series.boundary_fluxes[n].y_momentum;
+
+        // tab-delimited data format is:
+        // 1. simulation time (in desired units)
+        // 2. (global) index of boundary edge
+        // 3. integer ID indicating the boundary to which the boundary edge belongs
+        // 4. integer ID indicating the type of boundary condition on the boundary edge
+        // 5. water mass flux
+        // 6. x-momentum flux
+        // 7. y-momentum flux
+        // 8. boundary edge normal x component
+        // 9. boundary edge normal y component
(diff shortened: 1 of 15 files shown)


jeff-cohere:
Note: what's dumped to the file is not the time-integrated boundary flux, but the cumulative flux since the last dump. It's very easy to switch from one to the other.

bishtgautam (commented):
Can you edit the `.yaml` to turn on the output for BC fluxes?

src/time_series.c
```diff
@@ -0,0 +1,198 @@
...
+    // if not, accumulate fluxes locally
+    if (!auto_generated) {
+      PetscInt n = rdy->time_series.boundary_fluxes.offsets[b];
+      for (PetscInt e = 0; e < boundary->num_edges; ++e) {
+        PetscInt edge_id = boundary->edge_ids[e];
+        PetscInt cell_id = rdy->mesh.edges.cell_ids[2 * edge_id];
+        if (rdy->mesh.cells.is_local[cell_id]) {
+          time_series->boundary_fluxes.fluxes[n].water_mass += fluxes[e][0];
+          time_series->boundary_fluxes.fluxes[n].x_momentum += fluxes[e][1];
+          time_series->boundary_fluxes.fluxes[n].y_momentum += fluxes[e][2];
```
bishtgautam, line 67: I think this needs to be multiplied by `dt`.

src/time_series.c
```diff
@@ -0,0 +1,198 @@
...
+          local_flux_md[3 * n]     = rdy->mesh.edges.global_ids[edge_id];
+          local_flux_md[3 * n + 1] = rdy->boundary_ids[b];
+          local_flux_md[3 * n + 2] = bc.flow->type;
+
+          local_flux_data[5 * n]     = rdy->time_series.boundary_fluxes.fluxes[n].water_mass;
+          local_flux_data[5 * n + 1] = rdy->time_series.boundary_fluxes.fluxes[n].x_momentum;
+          local_flux_data[5 * n + 2] = rdy->time_series.boundary_fluxes.fluxes[n].y_momentum;
+          RDyVector edge_normal      = rdy->mesh.edges.normals[edge_id];
+          local_flux_data[5 * n + 3] = edge_normal.V[0];
+          local_flux_data[5 * n + 4] = edge_normal.V[1];
```
bishtgautam, line 107: The `3` and `5` could be defined as variables e.g., `PetscInt num_md = 3, num_data = 5;`

src/time_series.c
```diff
@@ -0,0 +1,198 @@
...
+  PetscInt n = 0;
+  for (PetscInt b = 0; b < rdy->num_boundaries; ++b) {
+    RDyCondition bc = rdy->boundary_conditions[b];
+    if (!bc.auto_generated) {  // exclude auto-generated boundary conditions
+      RDyBoundary *boundary = &rdy->boundaries[b];
+      for (PetscInt e = 0; e < boundary->num_edges; ++e) {
+        PetscInt edge_id = boundary->edge_ids[e];
+        PetscInt cell_id = rdy->mesh.edges.cell_ids[2 * edge_id];
+        if (rdy->mesh.cells.is_local[cell_id]) {
+          local_flux_md[3 * n]     = rdy->mesh.edges.global_ids[edge_id];
```
bishtgautam, line 98: I'm not certain if `rdy->mesh.edges.global_ids[edge_id];` has the correct ID when running in parallel. If it doesn't, the code should be fixed in another PR.

bishtgautam: I believe the objective here is to have the edge ID be in the natural order. Before distributing the mesh, we only have code that makes sure the natural ID of cells is known after the mesh is partitioned. I guess we need to add code before distributing the mesh, so the natural ID of vertices and edges is know after mesh partition.

src/time_series.c
jeff-cohere: The strides of 3 and 5 are encoded in variables now, but these values are still "hard-wired" in the sense that one still has to step over values 0-2 (lines 98-100 above) and 0-4 (lines 102-107 above) respectively.

src/time_series.c
bishtgautam: Opened #68

src/time_series.c
```diff
@@ -0,0 +1,201 @@
...
+    RDyCondition bc = rdy->boundary_conditions[b];
+    if (!bc.auto_generated) {
+      RDyBoundary *boundary = &rdy->boundaries[b];
+      for (PetscInt e = 0; e < boundary->num_edges; ++e) {
+        PetscInt edge_id = boundary->edge_ids[e];
+        PetscInt cell_id = rdy->mesh.edges.cell_ids[2 * edge_id];
+        if (rdy->mesh.cells.is_local[cell_id]) ++num_boundary_edges;
+      }
+    }
+    rdy->time_series.boundary_fluxes.offsets[b + 1] = 3 * num_boundary_edges;
```
bishtgautam, line 24: Let's get `3 (=ndof)` from PETSc 

```
  PetscInt ndof;
  PetscCall(VecGetBlockSize(rdy->X_local, &ndof));
```

src/time_series.c
```diff
@@ -0,0 +1,201 @@
...
+  }
+  rdy->time_series.boundary_fluxes.num_local_edges = num_boundary_edges;
+
+  // determine the global number of boundary edges
+  MPI_Allreduce(&num_boundary_edges, &rdy->time_series.boundary_fluxes.num_global_edges, 1, MPI_INT, MPI_SUM, rdy->comm);
+
+  // allocate (local) boundary flux storage
+  PetscCall(RDyAlloc(PetscReal, 3 * num_boundary_edges, &(rdy->time_series.boundary_fluxes.fluxes)));
```
bishtgautam, line 32: and use `ndof` here.

src/time_series.c
jeff-cohere: This is a little more complicated, because this `3` is actually hard-wired in various ways at the moment. Among them:

* The `rdy.time_series.boundary_fluxes` struct has three named fields that identify the boundary fluxes
* The `AccumulateBoundaryFluxes` function takes an array whose last dimension is `3` (the number of boundary fluxes), and changing that to a runtime parameter would be a little awkward (but doable)

We could check that `ndof == 3` above and leave a note in `rdycoreimpl.h` that the number of DOF must match the number of boundary fluxes. I don't really want to give the impression that everything will magically work when someone changes one of these numbers.

bishtgautam: Okay, let's leave `3` for now. When we will add the sediment dynamics based on @donghuix's https://github.com/RDycore/toy-problems/pull/22, we will automatically have to revisit the hard-wired value of `3` in multiple places.

bishtgautam (approved):

jeff-cohere:
@jedbrown , we're merging this PR and setting up an issue that lists all the things we're missing in the CEED implementation.

PR #70, opened 2023-08-02
Adds the Fortran interface for setting coupling interval

bishtgautam:
The Fortran interface for RDyGetCouplingInterval is added.

### DIFF
--- a/src/f90-mod/rdycore.F90
+++ b/src/f90-mod/rdycore.F90
@@ -80,7 +80,7 @@ module rdycore
     integer(c_int) function rdysetcouplinginterval_(rdy, interval) bind(c, name="RDySetCouplingInterval")
       use iso_c_binding, only: c_int, c_ptr, c_double
       type(c_ptr),    value, intent(in) :: rdy
-      real(c_double),        intent(in) :: interval
+      real(c_double), value, intent(in) :: interval
     end function
 
     integer(c_int) function rdygetheight_(rdy, h) bind(c, name="RDyGetHeight")
--- a/src/f90-mod/rdycore.F90
+++ b/src/f90-mod/rdycore.F90
@@ -77,6 +77,12 @@ module rdycore
       real(c_double),        intent(out) :: interval
     end function
 
+    integer(c_int) function rdysetcouplinginterval_(rdy, interval) bind(c, name="RDySetCouplingInterval")
+      use iso_c_binding, only: c_int, c_ptr, c_double
+      type(c_ptr),    value, intent(in) :: rdy
+      real(c_double),        intent(in) :: interval
+    end function
+
     integer(c_int) function rdygetheight_(rdy, h) bind(c, name="RDyGetHeight")
       use iso_c_binding, only: c_int, c_ptr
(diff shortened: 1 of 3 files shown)


src/f90-mod/rdycore.F90
```diff
@@ -77,6 +77,12 @@ integer(c_int) function rdygetcouplinginterval_(rdy, interval) bind(c, name="RDy
       real(c_double),        intent(out) :: interval
     end function
 
+    integer(c_int) function rdysetcouplinginterval_(rdy, interval) bind(c, name="RDySetCouplingInterval")
+      use iso_c_binding, only: c_int, c_ptr, c_double
+      type(c_ptr),    value, intent(in) :: rdy
+      real(c_double),        intent(in) :: interval
```
jeff-cohere, line 83: This parameter should have the `value` attribute too (otherwise it gets passed as a pointer).

bishtgautam: I believe I also made a mistake in the line above it also as `rdy` will be updated. So, the correction should be the following:

```
diff --git a/src/f90-mod/rdycore.F90 b/src/f90-mod/rdycore.F90
index efce6a2..bf4781c 100644
--- a/src/f90-mod/rdycore.F90
+++ b/src/f90-mod/rdycore.F90
@@ -79,8 +79,8 @@ module rdycore
 
     integer(c_int) function rdysetcouplinginterval_(rdy, interval) bind(c, name="RDySetCouplingInterval")
       use iso_c_binding, only: c_int, c_ptr, c_double
-      type(c_ptr),    value, intent(in) :: rdy
-      real(c_double),        intent(in) :: interval
+      type(c_ptr),           intent(in) :: rdy
+      real(c_double), value, intent(in) :: interval
     end function
```

Does that look good?

bishtgautam: Nope, the correction should be.

```
diff --git a/src/f90-mod/rdycore.F90 b/src/f90-mod/rdycore.F90
index efce6a2..deb706c 100644
--- a/src/f90-mod/rdycore.F90
+++ b/src/f90-mod/rdycore.F90
@@ -79,8 +79,8 @@ module rdycore
 
     integer(c_int) function rdysetcouplinginterval_(rdy, interval) bind(c, name="RDySetCouplingInterval")
       use iso_c_binding, only: c_int, c_ptr, c_double
-      type(c_ptr),    value, intent(in) :: rdy
-      real(c_double),        intent(in) :: interval
+      type(c_ptr),           intent(inout) :: rdy
+      real(c_double), value, intent(in) :: interval
     end function
 
```

jeff-cohere: Actually, `rdy` should have the `value` attribute too, so that its value (a C pointer) is passed properly. If you don't have the `value` attribute on a `c_ptr`, it gets passed as a pointer to that pointer.

jeff-cohere: It's a bit confusing--the data _pointed to by_ `rdy` changes, but the _value_ of `rdy` (its location in memory) doesn't change.

jeff-cohere (approved):

PR #76, opened 2023-08-22
Fixes an MPI hang in writing boundary fluxes when one or more subdomains has no boundary edges.

### LINKED ISSUE #74: The logic to output boundary flux is not correct for parallel runs
In a parallel run, a job hangs because of the following reason:
-  A MPI rank could have no boundary edges (i.e., `rdy->time_series.boundary_fluxes.fluxes`  is `0`) https://github.com/RDycore/RDycore/blob/main/src/time_series.c#L38
- In `WriteTimeSeries`, the `WriteBoundaryFluxes()` is only called for MPI ranks that have non-zero boundary edges https://github.com/RDycore/RDycore/blob/main/src/time_series.c#L196
- But, there is a `MPI_Gather` implemented within `WriteBoundaryFluxes()`

@jeff-cohere I have a `.yaml` and `.exo` file to reproduce this error that I will share with you.

jeff-cohere:
Reproducing this problem required a patch to PETSc, so we can't add the tests of interest to our suite just yet. But this fix, which makes all processes participate in the calls to `MPI_Gather` and `MPI_Gatherv` used to compute boundary fluxes, seems to resolve the issue that @bishtgautam encountered.

Closes #74

### DIFF
--- a/src/time_series.c
+++ b/src/time_series.c
@@ -87,12 +87,16 @@ static PetscErrorCode WriteBoundaryFluxes(RDy rdy, PetscInt step, PetscReal time
   PetscInt num_global_edges = rdy->time_series.boundary_fluxes.num_global_edges;
 
   // flux metadata (global edge ID, boundary ID, BC type)
-  int      num_md = 3;
-  PetscInt local_flux_md[num_md * num_local_edges];
+  // (add a padding byte to the end to prevent a 0-length VLA in case we don't
+  // have any local boundary edges)
+  PetscInt num_md = 3;
+  PetscInt local_flux_md[num_md * num_local_edges + 1];
 
   // flux data itself (mass, x-momentum, y-momentum, edge x normal, edge y normal)
-  int       num_data = 5;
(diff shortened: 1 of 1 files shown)


bishtgautam (approved):
@jeff-cohere Thanks for the fix and it is good to go.

Though, I believe [the first two](https://github.com/RDycore/RDycore/pull/76/files#diff-6b0daf976f2484e61288af70ce4f925e3f9516bf28936976b119bbfd2d84bc78R183-R184) `MPI_Gather` can be done once and the data can be saved. What do you think? If yes, this can be enchanced in a future PR.

jeff-cohere:
> Though, I believe [the first two](https://github.com/RDycore/RDycore/pull/76/files#diff-6b0daf976f2484e61288af70ce4f925e3f9516bf28936976b119bbfd2d84bc78R183-R184) `MPI_Gather` can be done once and the data can be saved. What do you think? If yes, this can be enchanced in a future PR.

That's a great idea. I'll create an issue for it and merge this PR.

PR #79, opened 2023-08-29
Fixes an issue with preloading, and removes some unnecessary code.

### LINKED ISSUE #78: Clean up preloading logic
Says @jedbrown, regarding our [preloading logic](https://github.com/RDycore/RDycore/blob/main/src/rdyadvance.c#L244-L252):

> ... I think the manual stuff (checking the option and outer conditional) can be removed since the macros and PetscPreLoadingOn handle that, doing what you want when -preload is not passed.

jeff-cohere:
@bishtgautam , let's see if this fixes the issue you're seeing with preloading in your performance measurements.

Closes #78


bishtgautam:
Unfortunately, it did not fix the issue. see https://gist.githubusercontent.com/bishtgautam/187b84b7bfe415ce4ca3f4323863d2bf/raw/14d5f233e8b6c8b71ed7e51131edd3b3cfa50389/gistfile1.txt

bishtgautam:
The libCEED example that @jedbrown has previously shared as a template of how preloading is used is at  https://github.com/CEED/libCEED/blob/main/examples/fluids/src/setupts.c#L502-L524

The profiling data of the libCEED example correctly shows information: see
https://gist.githubusercontent.com/bishtgautam/745d22acf5e9e1dcb78e64b5aec8028d/raw/cffc9accbe22c51c49dce729204e036cfbbf430c/gistfile1.txt

jeff-cohere:
Unfortunately Jed's example doesn't work for us in situations where we call `RDyAdvance` more than once (when we specify a `coupling_interval` that requires us to do so). The reason for this is that the preloading macros assume they exist in a function that is called exactly once. If they are invoked in a function that is called more than once, we get the following error:

```
[0]PETSC ERROR: Duplicate stage name given: RDyAdvance solve
```

The line of code that produces this error is:

```
PetscPreLoadBegin(PETSC_FALSE, "RDyAdvance solve");
```

This error occurs because the above macro attempts to register a log stage called "RDyAdvance solve" during the preload phase (or if preloading is disabled), assuming that it doesn't exist. And it does indeed exist the second or later time that `PetscPreLoadBegin` is called.

So we're violating the conditions under which these macros are assumed to operate, unless I'm misunderstanding something. @jedbrown , do you have any examples that allow a solver to be invoked more than once with preloading? I know this is not the "PETSc way" of doing things, but our need to couple with E3SM makes it difficult for us to surrender control of our time integration scheme to PETSc, and I'd prefer to avoid complicated solutions of the sort that can trick PETSc into thinking that it does actually have full control when it really doesn't.

All that said, I'm not sure I understand yet why my workaround isn't working, so I can keep picking at that.

jedbrown:
I think the PETSc macro is programmed defensively, but not tested in this regime (and thus not behaving as desired). I think we could skip the preload iteration when the stage has already been registered, and then the simple code in rdycore would work. I can look into this after class if desired.

jeff-cohere:
That would be great if it's practical for you, Jed!

bishtgautam:
@jedbrown Even if we can get the profiling information for only a single TSSolve problem that would be good enough for now.

jedbrown:
How would I test the multi-step mode with rdycore?

bishtgautam:
@jedbrown The multi-step mode example can be run via:

```
cd <your-build-dir/driver/tests/swe_roe>
../../rdycore ex2b_ic_file.yaml
```

jeff-cohere:
Thanks Jed! After your upstream merge request goes into PETSc, we can rebase this branch and try things out.

jedbrown:
It has merged to PETSc `release` (and will be in `main` soon, though I don't know which RDycore prefers).

jeff-cohere:
We've been tracking the `main` branch for the last few updates to our Docker images. It seems like we've been taking advantage of new changes , but maybe when things settle down a bit, we can take a closer look at releases again.

jeff-cohere:
Okay, I rebased this against our `main` branch and tidied some things up. The failing tests reflect the fact that the upstream merge isn't included in our Docker images yet. I think we can return to this after AGU and the holidays.

bishtgautam (approved):

PR #84, opened 2023-10-02
Allowing the log file to be overridden with a function call

### LINKED ISSUE #75: Ability to pass to a pointer of the logging file already created
For an E3SM-RDycore run, E3SM will open a logging file named `rdy.log.JOBID` (e.g. see `iulog` in https://github.com/RDycore/E3SM/commit/6f9963c5c302e7e86aad3d66def1eab6816412af#diff-f561dda7e4923546b71863e7959473d32a0668271a8ffe2b410ede9ede15e1c9R50). We need the ability to pass to RDycore `iulog`, so it can be set in `rdy->log` during [`RDySetup()`](https://github.com/RDycore/RDycore/blob/main/src/rdysetup.c#L931-L935)

jeff-cohere:
This PR adds an `RDySetLogFile` function in C and Fortran to allow an external client (e.g. E3SM) to override the log file specified in the input YAML file.

Closes #75

### DIFF
--- a/include/rdycore.h.in
+++ b/include/rdycore.h.in
@@ -26,6 +26,9 @@ PETSC_EXTERN PetscErrorCode RDyOnFinalize(void (*)(void));
 PETSC_EXTERN PetscErrorCode RDyFinalize(void);
 PETSC_EXTERN PetscBool RDyInitialized(void);
 
+// RDycore online configuration
+PETSC_EXTERN PetscErrorCode RDySetLogFile(const char*);
+
 // RDycore setup/breakdown
 PETSC_EXTERN PetscErrorCode RDyCreate(MPI_Comm, const char*, RDy*);
 PETSC_EXTERN PetscErrorCode RDySetup(RDy);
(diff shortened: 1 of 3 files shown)


jeff-cohere:
@bishtgautam , it may be a good idea for us to write a test program that behaves like E3SM would, in order to test our coupling strategy separately from the dycore's primary functions. This would allow us to differentiate between how we use our standalone drivers and how we imagine E3SM using RDycore. Let me know if you'd like to discuss.

bishtgautam (commented):
This is great. How about we also update the `main/driver/main.F90` do use this new function?

jeff-cohere:
> This is great. How about we also update the `main/driver/main.F90` do use this new function?

This will work, as long as we don't mind hardwiring the log file name, or accepting it as an input parameter. My idea above for writing a "coupler" test program would allow us to use the F90 driver as is, but either approach works for me.

bishtgautam:
Okay, let's leave the current driver as is and we will add the call to `RDySetLogFile` in the new "coupler" test program.

bishtgautam (approved):

PR #86, opened 2023-10-02
Setting Vec and Mat types appropriately when we use CEED.

### LINKED ISSUE #73: Set VecType MatType based on CEED backend memory
Follow the approach used in [fluids-ceed](https://github.com/CEED/libCEED/blob/main/examples/fluids/navierstokes.c#L133-L151) to set PETSc VecType and MatType. This will avoid the need explicitly specify `-dm_vec_type cuda` when using `-ceed /gpu/cuda`.

jeff-cohere:
Additionally, this PR gives the `RDy` type its own `ceed` so we don't have to keep creating CEED contexts wherever we need to use it.

Closes #73

### DIFF
--- a/src/physics_swe.c
+++ b/src/physics_swe.c
@@ -115,9 +115,7 @@ static PetscErrorCode RDyCeedOperatorSetUp(RDy rdy, PetscReal dt) {
   if (rdy->ceed_resource[0] && !rdy->ceed_rhs.op_edges) {
     rdy->ceed_rhs.dt = 0.0;
 
-    Ceed ceed;
-    CeedInit(rdy->ceed_resource, &ceed);
-    CeedCompositeOperatorCreate(ceed, &rdy->ceed_rhs.op_edges);
+    CeedCompositeOperatorCreate(rdy->ceed, &rdy->ceed_rhs.op_edges);
     CeedInt   num_comp = 3;
     RDyMesh  *mesh     = &rdy->mesh;
     RDyCells *cells    = &mesh->cells;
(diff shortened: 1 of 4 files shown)


jeff-cohere:
We probably want to test this with a CUDA build before we approve it.

src/rdysetup.c
```diff
@@ -123,6 +124,30 @@ static PetscErrorCode CreateDM(RDy rdy) {
...
+      case CEED_MEM_DEVICE: {
+        const char *resolved;
+        PetscCallCEED(CeedGetResource(rdy->ceed, &resolved));
+        if (strstr(resolved, "/gpu/cuda")) vec_type = VECCUDA;
+        else if (strstr(resolved, "/gpu/hip")) vec_type = VECKOKKOS;
+        else if (strstr(resolved, "/gpu/sycl")) vec_type = VECKOKKOS;
+        else vec_type = VECSTANDARD;
+      }
+    }
+    PetscCall(DMSetMatType(rdy->dm, mat_type));
```
bishtgautam, line 147: The `mat_type` should be also set like https://github.com/CEED/libCEED/blob/main/examples/fluids/navierstokes.c#L150C3-L152

jeff-cohere: Thanks, fixed!

src/physics_swe.c
```diff
@@ -115,25 +115,23 @@ static PetscErrorCode RDyCeedOperatorSetUp(RDy rdy, PetscReal dt) {
   if (rdy->ceed_resource[0] && !rdy->ceed_rhs.op_edges) {
     rdy->ceed_rhs.dt = 0.0;
 
-    Ceed ceed;
-    CeedInit(rdy->ceed_resource, &ceed);
-    CeedCompositeOperatorCreate(ceed, &rdy->ceed_rhs.op_edges);
+    CeedCompositeOperatorCreate(rdy->ceed, &rdy->ceed_rhs.op_edges);
```
bishtgautam, line 118: Ideally, `RDyCeedOperatorSetUp()` shouldn't be called within a time step and done during initialization. Fixing this would require moving the call to `RDyCeedOperatorSetUp` outside of `RDyCeedOperatorApply()`. Should this be fixed in this PR itself or in another PR?

jeff-cohere: I can move `RDyCeedOperatorSetup` outside of `RDyCeedOperatorApply` and see what happens.

bishtgautam (approved):

PR #88, opened 2023-10-06
Moves the setup of CEED Operator out of RHSFunction

bishtgautam:
The CEED Operator is now setup at the end of RDySetup instead of setting 
it up when RHSFunction is called for the first time.


src/swe_setup_ceed.h
```diff
@@ -0,0 +1,311 @@
+#ifndef swe_setup_ceed_h
+#define swe_setup_ceed_h
+
+#include "swe_flux_ceed.h"
+
+// Free a plain data context that was allocated using PETSc; returning libCEED error codes
+int FreeContextPetsc(void *data) {
```
jeff-cohere, line 7: This can be declared `static`, since it's not used anywhere outside this file.

src/swe_setup_ceed.h
```diff
@@ -0,0 +1,311 @@
...
+    CeedVectorCreate(rdy->ceed, mesh->num_cells_local * num_comp, &rdy->ceed_rhs.s_ceed);
+    CeedVectorCreate(rdy->ceed, mesh->num_cells_local * num_comp, &rdy->ceed_rhs.u_ceed);
+
+    if (0) CeedOperatorView(rdy->ceed_rhs.op_src, stdout);
+  }
+
+  PetscFunctionReturn(0);
+}
+
+#endif  // swe_setup_cee_h
```
jeff-cohere, line 311: little typo (but it hardly matters): `swe_setup_ceed_h`

jeff-cohere (approved):
Thanks, Gautam. In the light of performance analysis, this is likely a better solution than the one I used in my last PR.

PR #89, opened 2023-10-06
Reduces memory allocation/deallocating within each time step

bishtgautam:
In the PETSc version of the solver, few data structures are allocated during initialization 
for computing internal, boundary, and source term. This replaces the pervious code in
which data structures were allocated/deallocated within each time step.

### DIFF
--- a/include/private/rdycoreimpl.h
+++ b/include/private/rdycoreimpl.h
@@ -177,6 +177,9 @@ struct _p_RDy {
     CeedScalar           dt;
 
   } ceed_rhs;
+
+  RiemannDataSWE *datal_internal_edges, *datar_internal_edges;
+  RiemannDataSWE **datal_bnd_edges, **datar_bnd_edges;
 };
 
 PETSC_INTERN PetscErrorCode ReadConfigFile(RDy);
--- a/include/private/rdycoreimpl.h
+++ b/include/private/rdycoreimpl.h
@@ -6,6 +6,7 @@
 #include <private/rdyconfigimpl.h>
 #include <private/rdylogimpl.h>
 #include <private/rdymeshimpl.h>
+#include <private/rdysweimpl.h>
 #include <rdycore.h>
 
 // This type defines a region consisting of cells identified by their local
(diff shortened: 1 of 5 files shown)


bishtgautam:
The tests are failing because they aren't able to upload data to Codecov. :(

jeff-cohere:
> The tests are failing because they aren't able to upload data to Codecov. :(

Yeah, this happens occasionally. I've found rerunning the test job (sometimes a couple times, or at a later time) usually fixes it.

include/private/rdycoreimpl.h
```diff
@@ -176,6 +177,10 @@ struct _p_RDy {
     CeedScalar           dt;
 
   } ceed_rhs;
+
+  RiemannDataSWE datal_internal_edges, datar_internal_edges;
```
jeff-cohere, line 181: If you think it's helpful and not too much typing, consider placing these into a struct variable like `ceed_rhs` above (e.g. named `data_swe` or `swe_fluxes`).

src/rdysetup.c
```diff
@@ -967,6 +967,59 @@ PetscErrorCode RDySetLogFile(RDy rdy, const char *filename) {
   PetscFunctionReturn(0);
 }
 
+/// For computing fluxes, allocates structs to hold values left and right
+/// of internal and boundary edges
+PetscErrorCode InitDataStructsForComputingFlux(RDy rdy) {
```
jeff-cohere, line 972: We could call this `AllocateFluxStorage` (just another option)

src/rdysetup.c
```diff
@@ -967,6 +967,59 @@ PetscErrorCode RDySetLogFile(RDy rdy, const char *filename) {
...
+  }
+
+  rdy->datal_bnd_edges = &(*datal_bnd);
+  rdy->datar_bnd_edges = &(*datar_bnd);
+
+  PetscFunctionReturn(0);
+}
+
+/// Allocates a struct for computing the source/sink term
+PetscErrorCode InitDataStructsForSourceTerm(RDy rdy) {
```
jeff-cohere, line 1007: Similarly, we might call this `AllocateSourceTermStorage`

jeff-cohere (approved):
This looks great! I left some suggestions for you to consider, but I'm okay merging as is.

PR #90, opened 2023-10-12
Ceed operator flux accumulation part 1: flux storage, code reorganization

jeff-cohere:
This PR

* adds an accumulated flux (on edges) as a passive output to the existing edge (flux) operator
* moves SWE operator code to its own `swe` folder
* defines a simplified interface for creating flux and source operators
* defines a pair of functions for extracting CEED fields from a source operator
* adds an `id` field to the `RDyBoundary` stuct and removes the `boundary_ids` field from `RDy`
* removes some unnecessary parameters

This seems like a good point of departure for implementing our more elaborate coupling features. I envision another function for extracting accumulated (edge) fluxes from the flux operator, for example.


jeff-cohere:
Grrr, the GitHub Actions thing we use for Clang formatting no longer supports `clang-format` version 14, which is our pegged version. I suppose maybe it's time to update to another version? Or we can get rid of this externality and do our own format checking.

bishtgautam:
Let's check what is libceed doing for automatic format testing?

jeff-cohere:
libCEED seems to be using `clang-format-16`: https://github.com/CEED/libCEED/blob/main/.github/workflows/c-fortran-test-style.yml

jeff-cohere:
Aha, it's possible that I've misinterpreted the format check error. Let me try something.

jeff-cohere:
Okay, fixed! Sorry about the grumbling.

src/swe/swe_operators.c
```diff
@@ -0,0 +1,364 @@
...
+  CeedQFunctionContextSetDataDestroy(*qf_context, CEED_MEM_HOST, FreeContextPetsc);
+  CeedQFunctionContextRegisterDouble(*qf_context, "time step", offsetof(struct SWEContext_, dtime), 1, "Time step of TS");
+  CeedQFunctionContextRegisterDouble(*qf_context, "small h value", offsetof(struct SWEContext_, tiny_h), 1,
+                                     "Height threshold below which dry condition is assumed");
+  CeedQFunctionContextRegisterDouble(*qf_context, "gravity", offsetof(struct SWEContext_, gravity), 1, "Accelaration due to gravity");
+
+  PetscFunctionReturn(0);
+}
+
+PetscErrorCode CreateSWEFluxOperator(Ceed ceed, RDyMesh *mesh, int num_boundaries, RDyBoundary boundaries[num_boundaries],
```
bishtgautam, line 34: How about splitting this long function into two (e.g., `CreateSubOpForInternalEdges` and `CreateSubOpForBoundaryEdges`)?

src/swe/swe_operators.c
```diff
@@ -0,0 +1,364 @@
...
+
+    {
+      CeedOperator op;
+      CeedOperatorCreate(ceed, qf, NULL, NULL, &op);
+      CeedOperatorSetField(op, "geom", restrict_geom, CEED_BASIS_COLLOCATED, geom);
+      CeedOperatorSetField(op, "q_left", restrict_l, CEED_BASIS_COLLOCATED, CEED_VECTOR_ACTIVE);
+      CeedOperatorSetField(op, "q_right", restrict_r, CEED_BASIS_COLLOCATED, CEED_VECTOR_ACTIVE);
+      CeedOperatorSetField(op, "cell_left", restrict_l, CEED_BASIS_COLLOCATED, CEED_VECTOR_ACTIVE);
+      CeedOperatorSetField(op, "cell_right", restrict_r, CEED_BASIS_COLLOCATED, CEED_VECTOR_ACTIVE);
+      CeedOperatorSetField(op, "flux", restrict_flux, CEED_BASIS_COLLOCATED, flux);
```
bishtgautam, line 119: Since we specified a vector (`flux` as the fifth argument) instead of `CEED_VECTOR_ACTIVE`, it becomes a "passive field". The `geom` is also a passive field except `geom` is an input field (via [`CeedQFunctionAddInput#L47`](https://github.com/RDycore/RDycore/pull/90/files#diff-cebc7026ff0766eb019ae159a8319ca18c5fa7d87b58278c9ed9ec7957ac2e53R47)), while `flux` is an output field (via [`CeedQFunctionAddOutput#L52`](https://github.com/RDycore/RDycore/pull/90/files#diff-cebc7026ff0766eb019ae159a8319ca18c5fa7d87b58278c9ed9ec7957ac2e53R52)). Did I get this correct?

bishtgautam (commented):
This looks great. I have two comments:
- a minor comment about splitting a bigger function into two, and 
- Should we add a code block that is the reverse of [`if (!rdy->ceed_rhs.water_src_updated)`](https://github.com/RDycore/RDycore/pull/90/files#diff-08c17314ab10e0cbbeec2a48b2c6c6b69e0ad40f647b8fc64edb99f717f6d33cL108-L138) to copy fluxes from a `CeedVector` to a `Vec`? Or, should this code be added in another PR?

src/swe/swe_operators.c
jeff-cohere: That's a good idea. I'll break it up.

src/swe/swe_operators.c
jeff-cohere: Yes, that is how I understand it.

jeff-cohere:
>     * Should we add a code block that is the reverse of [`if (!rdy->ceed_rhs.water_src_updated)`](https://github.com/RDycore/RDycore/pull/90/files#diff-08c17314ab10e0cbbeec2a48b2c6c6b69e0ad40f647b8fc64edb99f717f6d33cL108-L138) to copy fluxes from a `CeedVector` to a `Vec`? Or, should this code be added in another PR?

I think copying the fluxes into a `Vec` so we can use them for coupling is the next step, but I think we might like to discuss how we plan to do this coupling in more detail before we determine exactly how this works. For example, we currently just write the boundary fluxes out to a file in the PETSc flux accumulation scheme, but we might want to do a direct copy of these fluxes to E3SM/ELM instead of using file coupling. So I've left those decisions for a "part 2" PR.

Does this make sense? Am I missing anything?

bishtgautam:
Okay, I agree that we should the copy of fluxes from `CeedVector` to `Vec` in a future PR.

jeff-cohere:
@bishtgautam, for safety, maybe we should test that these changes don't mess anything up in a GPU setting. I moved some logic that was in a header file into a `.c` file, and I'm not familiar enough with the JIT machinery that CEED uses to understand whether this change in "visibility" makes a difference.

jedbrown:
The JIT expects to be able to include the header file, so if you have logic called by a qfunction (mostly those using `CEED_QFUNCTION_HELPER`, or macros), it needs to be in the header. Everything else (in my opinion) should stay in `.c` files.

jeff-cohere:
That makes perfect sense to me. What we have here should be fine, then.

bishtgautam:
I verified that this PR is working correctly on Perlmutter GPU nodes.

bishtgautam (approved):

PR #91, opened 2023-10-14
This PR does some additional code cleanup.

jeff-cohere:
I think these changes are helpful, but I'm curious what others think.

* Moved swe_flux_petsc.h -> swe_flux_petsc.c
* Moved swe_operators.h function decls into private/rdysweimpl.h
* Removed SWE-specific code from RDy struct, replacing with an opaque pointer named petsc_rhs (analogous to ceed_rhs).
* Renamed swe/swe_operators.c -> swe/swe_ceed_operators.c for clarity.


jeff-cohere:
This PR is based off of #90, but will "move into position" when we merge that one.

bishtgautam (approved):
Looks great.

Issue #83, opened 2023-10-02
Implement boundary flux accumulation in libCEED implementation


jeff-cohere:
Currently, we can compute boundary fluxes using our PETSc implementation of the Riemann solver, but we also need this in a GPU setting with libCEED because this is a significant mechanism for coupling with E3SM.

It's probably easiest for us to formulate the boundary fluxes in terms of their contributions to neighboring cells so we can make use of CEED's work arrays on the same space we're using for the solution.

UPDATE: @bishtgautam and I discussed this task earlier today. The summary of our discussion is:

* We can accumulate fluxes in a stride-9 vector (3 DoF x 3 faces) for each cell, and designate this vector as a passive input/output during the RDycore integration, and then we can do whatever we need to do with this info when we do coupling data transfer at the end of the coupling timestep (or whenever)
* While we could (in principle, at least) restrict the accumulation of fluxes to boundary cells, we are going to do the accumulation on all cells (including interior cells) in case we want to plot velocity/flow vectors.

bishtgautam:
After our discussion, I realized that we should generalize the stride of the vector to be 3 DOFs x Max_faces_of_a_cell to support meshes comprising of triangles, quads, or both cell types.

jedbrown:
This sounds like a pretty heavy vector and I'm not clear on why it's needed, versus either ghost cells or a face operator that populates the external (ghost) cell state and solves its Riemann problem.

bishtgautam:
We **_need_** to accumulate fluxes going through boundary faces over multiple time steps. So, the approach that we came up with is to save the fluxes through edges in a cell-strided vector. In the current approach, we will be storing a duplicate copy of each flux. While we are adding this infrastructure for boundary fluxes, I thought we should also save the internal fluxes.

jeff-cohere:
@jedbrown , I didn't realize that we could define a "face" (edge) operator. That would probably be a better way to attack this feature, since it expresses the right relationship between edges and adjoining cells. Do you have any examples of face operators in existing codes?

jedbrown:
The current Riemann flux operator is on edges. Any such operator can store the result on the edges or accumulate it into adjacent cells.

jeff-cohere:
So it is! If there's a way to store the resulting fluxes on edges and subsequently use these fluxes to compute the adjacent cell values (via another operator, perhaps), I think we have what we need.

jedbrown:
[...]

PR #94, opened 2023-10-19
Implemented Dirichlet BCs with CEED and cleaned up water source updates

### LINKED ISSUE #95: Add CEED implementation to RDySetWaterSource
See [this comment](https://github.com/RDycore/RDycore/pull/94#discussion_r1366175317).

jeff-cohere:
This PR also renames some files and functions to provide slightly shorter and clearer names. It also moves some SWE-related code into the `swe` folder.

We discussed this item in #83, which is more focused on boundary flux accumulation, but related.

Closes #95


src/rdydata.c
```diff
@@ -23,60 +23,61 @@ static PetscErrorCode CheckBoundaryConditionID(RDy rdy, PetscInt bnd_cond_id) {
...
+    CeedOperatorFieldGetVector(dirichlet_field, &dirichlet_vector);
+    PetscInt num_comp = 3;
+    CeedScalar(*dirichlet_ceed)[num_comp];
+    CeedVectorGetArray(dirichlet_vector, CEED_MEM_HOST, (CeedScalar **)&dirichlet_ceed);
+
+    // set the boundary values
+    for (PetscInt i = 0; i < boundary.num_edges; ++i) {
+      dirichlet_ceed[i][0] = bc_values[num_comp * i];
+      dirichlet_ceed[i][1] = bc_values[num_comp * i + 1];
+      dirichlet_ceed[i][2] = bc_values[num_comp * i + 2];
```
bishtgautam, line 74: This is pretty neat and I think we should implement a similar code for [RDySetWaterSource()](https://github.com/RDycore/RDycore/blob/93285b4303002a000f127e81310197c28f9771b0/src/rdydata.c#L121-L131)

jeff-cohere:
Hmmm, it looks like we're not actually setting DIrichlet boundary values anywhere. I'll try to patch that in.

src/rdydata.c
jeff-cohere: Yeah, I was thinking the same thing. It makes our interface a bit smaller. I could do that in this PR, in fact, after I make sure we're actually testing the code that sets the Dirichlet BC values.

src/swe/swe_ceed.c
```diff
@@ -263,6 +291,45 @@ PetscErrorCode CreateSWEFluxOperator(Ceed ceed, RDyMesh *mesh, int num_boundarie
...
+  CeedCompositeOperatorGetSubList(flux_op, &sub_ops);
+  CeedOperator boundary_flux_op = sub_ops[1 + boundary.index];
+
+  // fetch the field
+  CeedOperatorGetFieldByName(boundary_flux_op, "q_dirichlet", boundary_values);
+
+  PetscFunctionReturn(0);
+}
+
+PetscErrorCode SWEFluxOperatorSetDirichletBoundaryValues(CeedOperator flux_op, RDyBoundary boundary, PetscReal *boundary_values) {
```
bishtgautam, line 309: Should we also send `num_comp` and `num_edges` as input arguments so we know the size of `*boundary_values` AND an `assert` to check the size of  `*boundary_values`?

jeff-cohere: Sure. This is a matter of taste, but as you point out it may help us find errors and inconsistencies.

bishtgautam: Something similar to https://github.com/RDycore/RDycore/pull/94/files#diff-2380356150436250abd03a1522da01310c2e79d6104107b17cb4667e1d723fe4R41

bishtgautam (approved):

src/swe/swe_ceed.c
bishtgautam:
```
PetscErrorCode SWEFluxOperatorSetDirichletBoundaryValues(CeedOperator flux_op, RDyBoundary boundary, PetscReal boundary_values[boundary.num_edges * ndof])
```

jeff-cohere: I think it's okay to handle these checks in `RDySetDirichletBoundaryValues` for now, but if we start to see these sorts of bugs crop up in our internal code, we can always add these parameters.

PR #97, opened 2023-10-24
Updates the C and Fortran drivers to run the Houston Harvey Flooding

bishtgautam:
Adds two command line arguments (`-bc` and `-rain`) to specify spatially homogenous, time varying boundary and rainfall condition to run the Houston Harvey flooding case.

```
./rdycore <yaml> \
-bc   <boundary-condition-file> \
-rain <rainfall-file>
```


bishtgautam:
The new tests are failing because there is a bug in PETSc for Exodus II that needs to be fixed.

driver/main.F90
```diff
@@ -36,6 +102,18 @@ program rdycore_f90
     PetscCallA(RDyInit(ierr))
 
     if (trim(config_file) /= trim('-help')) then
+
+      PetscCallA(PetscOptionsGetString(PETSC_NULL_OPTIONS, PETSC_NULL_CHARACTER, '-rain', rainfile, rain_specified, ierr))
+      PetscCallA(PetscOptionsGetString(PETSC_NULL_OPTIONS, PETSC_NULL_CHARACTER, '-bc', bcfile, bc_specified, ierr))
+
+      if (rain_specified) then
+        call opendata(rainfile, rain_vec, nrain)
```
jeff-cohere, line 110: Should we be calling `VecGetArrayF90` on `rain_ptr` here? And on `bc_ptr` below?

jeff-cohere (changes requested):
This looks good to me. I just pointed out one place in the Fortran driver in which pointers should be set.

jeff-cohere (approved):

PR #103, opened 2023-10-25
Finished implementation of CEED accumulated boundary fluxes

### LINKED ISSUE #83: Implement boundary flux accumulation in libCEED implementation

jeff-cohere:
This PR cleans up some boundary flux logic, adds a few comments to distinguish "fluxes" from "flux densities", and performs the device-to-host copy of the accumulated boundary fluxes from the CEED operators.

Closes #83

### DIFF
--- a/include/private/rdycoreimpl.h
+++ b/include/private/rdycoreimpl.h
@@ -192,7 +192,7 @@ PETSC_INTERN PetscErrorCode WriteXDMFOutput(TS, PetscInt, PetscReal, Vec, void *
 
 // time series
 PETSC_INTERN PetscErrorCode InitTimeSeries(RDy);
-PETSC_INTERN PetscErrorCode AccumulateBoundaryFluxes(RDy, RDyBoundary *, PetscInt ne, PetscReal[ne][3]);
+PETSC_INTERN PetscErrorCode AccumulateBoundaryFluxes(RDy, RDyBoundary boundary, PetscReal[boundary.num_edges][3]);
 PETSC_INTERN PetscErrorCode WriteTimeSeries(TS, PetscInt, PetscReal, Vec, void *);
 PETSC_INTERN PetscErrorCode DestroyTimeSeries(RDy);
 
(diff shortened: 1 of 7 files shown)


jeff-cohere:
@bishtgautam , I think we now have the same capabilities with CEED as we do with PETSc. We still need to figure out a way to copy the boundary fluxes to E3SM, but I think that should be easy once we establish an indexing scheme for boundary edges on the E3SM side.

bishtgautam (changes requested):
- In the CEED functions, need to multiply by `dt` at https://github.com/RDycore/RDycore/blob/main/src/swe/swe_ceed_impl.h#L125
- In order to use `AccumulateBoundaryFluxes` for PETSc and CEED versions, need to get rid of multiplication by `rdy->dt` in https://github.com/RDycore/RDycore/pull/103/files#diff-6b0daf976f2484e61288af70ce4f925e3f9516bf28936976b119bbfd2d84bc78R73
- Lastly, update the functions for PETSc version to multiply by `dt` so `AccumulateBoundaryFluxes()` can be reused by PETSc and CEED version of the solver.

bishtgautam (approved):

PR #113, opened 2023-11-23
Corrects the setting of values in a CeedVector

bishtgautam:
The size of the CeedVector associated with the boundary condition if 
of the total number of grid cells (local + ghosted). So, the dirichlet values
(p/hu/hv) need to be saved in the CeedVector corresponding to the `cell_id`
of boundary cells.


bishtgautam:
Corresponding discussion on libCeed is at https://github.com/CEED/libCEED/issues/1403

jeff-cohere (approved):
Looks great! Thanks for changing the variable names and everything.

PR #117, opened 2023-12-06
Updated Docker scripts to build int32 and int64 PETSc images.

### LINKED ISSUE #116: Add a new Docker image that uses 64-bit PETSc integer indices
We've recently started using 64-bit integer indices now that we're running on large meshes. @jedbrown has an [open PETSc merge request](https://gitlab.com/petsc/petsc/-/merge_requests/7076) to make libCEED play nicely under these circumstances. When that merge request goes into PETSc's main branch, we should create new Docker images for PETSc built with both 32- and 64-bit integer indices to allow us to build and test under both configurations.

jeff-cohere:
This PR includes updates to PETSc that allow libCEED to be built with 64-bit integer support, and modifies the Docker machinery to build two images (int32 and int64) for our CI testing.

Closes #116


jeff-cohere:
Excellent! The CI build fails with the 64-bit integer build of PETSc, so we know it's definitely different from the 32-bit integer version. I'll chase down these issues.

jeff-cohere:
I've successfully reproduced many of the test failures locally, but it looks like it might take some time for me to find out what's going on. Will keep this PR current with my progress.

jeff-cohere:
Okay, it looks to me like the remaining test errors result in tests that try to read data from binary PETSc files. I've attached an example of how the errors look at the bottom of this comment.

Variable-sized integer indices pose a problem for reading these binary data files. For example, any files written beforehand by RDycore using 32-bit indices will not be read properly by an RDycore build with 64-bit indices. For now, we could get around this problem by disabling the tests that rely on these pre-generated data files. Alternatively, we could create versions of these data files for both 32- and 64-bit indices. Then we'd just have to figure out which version of the data file to use within our build system.

@bishtgautam , do you have any thoughts on this?

```
[0]PETSC ERROR: --------------------- Error Message --------------------------------------------------------------
[0]PETSC ERROR: Unexpected data in file
[0]PETSC ERROR: Not a vector next in file
[0]PETSC ERROR: See https://petsc.org/release/faq/ for trouble shooting.
[0]PETSC ERROR: Abort(79) on node 1 (rank 0 in comm 16): application called MPI_Abort(MPI_COMM_SELF, 79) - process 0
Petsc Development GIT revision: v3.20.2-260-g713b55d3e87  GIT Date: 2023-12-06 17:10:02 +0000
[0]PETSC ERROR: /__w/RDycore/RDycore/build/driver/tests/../rdycore_f90 on a mpich-int64-real-debug named 6b697bfe0835 by Unknown Wed Dec  6 21:56:16 2023
[0]PETSC ERROR: Configure options --with-cc=mpicc --with-cxx=mpic++ --with-fc=mpif90 --CFLAGS="-g -O0" --CXXFLAGS="-g -O0" --FFLAGS="-g -O0 -Wno-unused-function" --with-clanguage=c --with-debugging=1 --with-shared-libraries=1 --download-exodusii --download-fblaslapack --download-hdf5 --download-cgns --download-libceed --download-metis --download-netcdf --download-parmetis --download-pnetcdf --with-zlib --with-64-bit-indices
[0]PETSC ERROR: #1 VecLoad_Binary() at /build/petsc/src/vec/vec/utils/vecio.c:80
[0]PETSC ERROR: #2 VecLoad_Default() at /build/petsc/src/vec/vec/utils/vecio.c:225
[0]PETSC ERROR: #3 VecLoad() at /build/petsc/src/vec/vec/interface/vector.c:1157
[0]PETSC ERROR: #4 ReadOneDOFVecFromFile() at /__w/RDycore/RDycore/src/rdysetup.c:603
[0]PETSC ERROR: #5 InitMaterials() at /__w/RDycore/RDycore/src/rdysetup.c:670
[0]PETSC ERROR: #6 RDySetup() at /__w/RDycore/RDycore/src/rdysetup.c:1051
```

jeff-cohere:
Okay, we've got two remaining tests that fail for reasons I don't yet understand. My latest commit introduces the ability to do some elementary variable substitution in input files so that we can avoid having to worry about getting the right data files. Specifically, if you put `${PETSC_ID_TYPE}` anywhere in your YAML file, it will be converted to `int32` or `int64` based on whether PETSc has been built with 64-bit indices. By adopting a simple convention for data files with `int32.bin` or `int64.bin` suffixes, we can use similar syntax in our CMake and YAML files for tests.

This is a step in the direction of code complexity, but it does allow us to sidestep some obvious workflow problems. If we like this solution, we're very close to merging this PR.

bishtgautam:
The testing using 64-bit integers will force us to be careful about using `integer`, `PetscInt`, `CeedInt` or `PetscMPIInt`

PR #121, opened 2024-01-08
Adds C and Fortran MMS drivers

bishtgautam:
- Adds a Fortran and C MMS driver.
- The two drivers can be run via:

```
cd <build-dir>/driver/tests/swe_roe

# Fortran driver
../../rdycore_mms_f90 mms_dx1.yaml
 Avg-cell-area    :  0.25000014920541758
 Avg-length-scale :  0.50000014920539526
 Error-Norm-1     :   9.9962646736427591E-003   1.5251880380016453E-003   1.5251880380016453E-003
 Error-Norm-2     :   2.4488131544836331E-003   4.4589086177372004E-004   4.4589086177372004E-004
 Error-Norm-Max   :   9.6211166633113077E-004   2.0563147306039691E-004   2.0563147306039685E-004

# C driver
../../rdycore_mms mms_dx1.yaml
Avg-cell-area    : 0.2500001492054176
Avg-length-scale : 0.5000001492053953
Error-Norm-1     : 0.0099962646736428 0.0015251880380016 0.0015251880380016
Error-Norm-2     : 0.0024488131544836 0.0004458908617737 0.0004458908617737
Error-Norm-Max   : 0.0009621116663311 0.0002056314730604 0.0002056314730604
```


bishtgautam:
@jeff-cohere The code is correctly compiling for 32 and 64bit on Mac and Ubuntu for me. Can you check if it compiles for you too? If it does compile correctly for you, any ideas on how to fix the errors in Github Actions?

jeff-cohere:
It compiles for me locally too. I think this is probably a bug in GFortran 11.2, which is what we're using in our CI environment. This week I can try to find a Docker configuration with newer compilers and see if it fixes this issue.

bishtgautam:
Thanks @jeff-cohere.

In that case, the PR is also ready for your review once the Docker is fixed.

jeff-cohere:
Curious. I've verified that my Linux workstation is using the same version of `gfortran` as this PR (11.4), but it works on that machine. I wonder if there's some strange race condition occurring in the build. I don't think a new Docker image would necessarily fix this.

I'm going to remove the multi-threaded build so we get a clearer look at the errors. That may shed some light.

jeff-cohere:
Apparently the issue is related to the legacy MPI Fortran bindings (`mpif.h`). As is explained [here](https://stackoverflow.com/questions/69686755/fortran-error-type-mismatch-between-actual-argument-at-1-and-actual-argument), a workaround is to add `-fallow-argument-mismatch` to the Fortran compiler's list of flags. I'll look over this issue now.

driver/tests/swe_roe/mms_dx1.yaml
```diff
@@ -0,0 +1,62 @@
+physics:
```
jeff-cohere, line 1: This is a good start. We discussed perhaps making another YAML file format for the MMS driver that would allow us greater flexibility in configuring the problem of interest, right? Is that still interesting?

driver/mms.c
```diff
@@ -0,0 +1,430 @@
...
+
+  if (strcmp(argv[1], "-help")) {  // if given a config file
+
+    // create rdycore and set it up with the given file
+    MPI_Comm comm = PETSC_COMM_WORLD;
+    RDy      rdy;
+
+    PetscCall(RDyCreate(comm, argv[1], &rdy));
+
+    PetscCall(RDySetup(rdy));
```
jeff-cohere, line 219: If we did make an alternate YAML format for MMS runs, we could add another version of this command (called `RDySetupMMS` or something) that would use a different (but similarly structured) input parser.

jeff-cohere (approved):
This is a great start on MMS and looks good to me. I left some minor comments about how we might add an alternate YAML input format that is tailored to MMS cases of interest.

driver/tests/swe_roe/mms_dx1.yaml
bishtgautam: As discussed during the call, the additional flexibility will be added in a future PR.

PR #123, opened 2024-01-16
Boundaries and boundary conditions are now constructed after domain decomposition

### LINKED ISSUE #122: Internal edges corresponding to ghost cells are getting added in the boundary data structure
Shown below is an example of grid decomposition on 2 MPI ranks for a problem that can be run as follows:

```
cd driver/tests/swe_roe
/opt/local/bin/mpiexec-openmpi-mp -n 2 ../../rdycore_mms mms_dx1.yaml
```

fullmesh
rank0
rank1

If a boundary condition is not specified on an edge, RDycore automatically adds a reflective boundary condition during model setup in [InitBoundaries()](https://github.com/RDycore/RDycore/blob/main/src/rdysetup.c#L440C23-L440C37). When running in parallel, there are edges of the red ghost cells that are identified as "boundary" cells ([`num_unassigned_edges`](https://github.com/RDycore/RDycore/blob/main/src/rdysetup.c#L510)). Since these are ghost cells, RDycore shouldn't be considering them as unassigned edges.

Currently, the `InitBoundaries()` is called before `RDyMeshCreateFromDM()`. Thus, we don't know if the cells corresponding to `num_unassigned_edges` are internal or ghost cells when `InitBoundaries()` is called.

A possible fix could be that as a last step of setup, we revisit the boundaries and check if the last boundary condition is of a reflecting type and remove all edges from the last boundary condition that correspond to ghost cells. If after removing all edges corresponding to ghost cells, there are no edges in that boundary, we should reduce `num_boundaries` by one.

jeff-cohere:
This is a very simple possible fix for #122. @bishtgautam , can you try this out with your setup and see if it behaves as it should?

Fixes #122

### DIFF
--- a/src/rdysetup.c
+++ b/src/rdysetup.c
@@ -449,9 +449,50 @@ static PetscErrorCode InitBoundaries(RDy rdy) {
 
   // Keep track of whether edges on the domain boundary have been assigned to
   // any boundaries.
-  IS unassigned_edges_is;
+  PetscInt e_start, e_end;  // starting and ending edge points
+  DMPlexGetHeightStratum(rdy->dm, 1, &e_start, &e_end);
+
+  IS       unassigned_edges_is;
+  PetscInt num_edges_invalid = 0;
+
+  // First remove all edges that are not valid. The invalid edges are:
+  // - edges between two non-local(=ghost) cells, and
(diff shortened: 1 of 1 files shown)


bishtgautam (approved):

jeff-cohere:
Thanks for the fix!

PR #124, opened 2024-01-17
Fixes application of BC for CEED.

bishtgautam:
- The size of the CeedVector to hold BC is changed from num_cells to num_edges.
- The CeedRestrict is updated to be consistent with the CeedVector size.
- Lastly, the index of the `dirichlet_ceed` in which the BC values are set is also updated
  to be consistent with the CeedVector size.

### DIFF
--- a/.github/workflows/clang-format-check.yml
+++ b/.github/workflows/clang-format-check.yml
@@ -10,4 +10,4 @@ jobs:
       uses: jidicula/clang-format-action@v4.11.0
       with:
         clang-format-version: '14'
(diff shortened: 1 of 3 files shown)


bishtgautam:
It turns out that fixing the BC was relatively straightforward. https://github.com/RDycore/RDycore/pull/124/files#diff-c3a185b2fadbb0e6b4e9bc0014b9c35d654022fd86f4ceee6ca26644a5a51738

jeff-cohere (approved):
Excellent!

PR #133, opened 2024-02-21
Fixes the support of mixed element type meshes in parallel

bishtgautam:
Terminology about PETSc Vecs
- A global PETSc Vec includes values for only local cells.
- A local PETSc Vec includes values for local and ghost cells.

For the different mesh types, the following is observed.
- Meshes with homogeneous element types: the ghost/halo cells are at the end of local cells.
- Meshes with mixed element types: the ghost/halo cells can be intermixed between local cells.

The RHSFunction receives a global solution Vec `X` and global flux Vec (`F`). The Vec `X` is scattered
to obtain a local Vec `X_local` via `DMGlobalToLocalBegin/End`. Then, the function fills out the local Vec `F`.

This PR adds two maps in the mesh:
1. `local_to_owned` (= G2L in the schematic) and
2. `owned_to_local` (= L2G in the schematic)

- For the PETSc code, the indices used to fill `F` use the `G2L` map. (e.g. [here])(https://github.com/RDycore/RDycore/pull/133/files#diff-40dfcd1e660eea33df6e70a1c8ec95e2d5acb31a34668332e2d3343f3837dffdR277-R285)

- For the CEED-version, the `CeedElemRestriction` are now created differently for the 
  inputs and outputs of CeedQFunction. The inputs use the ghosted index (same as before), 
  while the outputs use the local index via `G2L`. (see [here](https://github.com/RDycore/RDycore/pull/133/files#diff-c3a185b2fadbb0e6b4e9bc0014b9c35d654022fd86f4ceee6ca26644a5a51738R86-R91))

IMG_20240221_121523


jeff-cohere (approved):
Looks good to me.

jeff-cohere:
How about the following nomenclature:

* `local_to_rhs` instead of `G2L`
* `rhs_to_local` instead of `L2G`

This reminds us that we're mapping between the local and RHS vector cell indices.

bishtgautam:
If we implicit time integration, then we will code up a Jacobian. The values will be entered in the Jacobian in `G2L` (= `local_to_rhs`). IMIO, in such a case, the use of *rhs* will look odd.

jeff-cohere:
> If we implement implicit time integration, then we will code up a Jacobian. The values will be entered in the Jacobian in `G2L` (= `local_to_rhs`). IMIO, in such a case, the use of _rhs_ will look odd.

True. We could use `local_to_solver` or `local_to_soln` or something, to indicate that we're transforming to and from the vector space used by the solver.

jedbrown:
First, we can build a matrix using coloring (`-snes_fd_color`; no code). We can also compare fully matrix-free. We'll likely want to assemble a point-block diagonal (for `-pc_type pbjacobi` or `vpbjacobi`), but that can be approximate. I don't envision assembling the complete Jacobian, certainly not for the current operator (versus a lagged one used to build a lagged preconditioner). In any case, libCEED/PETSc have an interface to do this when we write the matrix-free action in a qfunction. That matrix-free action could be computed using Enzyme.

Issue #41, opened 2023-05-16
Add restart capability


jeff-cohere:
It should be possible for us to support restarts given an HDF5 file. We haven't really talked about this yet, so this issue is where we should have this conversation.

jedbrown:
I think @knepley has done this with the plex/xdmf schema for HDF5, though it likely needs some flags to get all the information in the file. I'm working on a CGNS parallel reader with restart capability (including labeled boundaries, etc.).

knepley:
Yes, this is now working as the standard way to do it in Firedrake. The complication is that you write out all the numberings, so that you can load on a different number of processes, but usually that is a lot smaller than the data.

bishtgautam:
In the RDycore, we create the DM with the natural ordering and then output the data in the natural order. So, I believe, unlike Friedrake, we don't have to save the numbering with the data.

knepley:
RDycore also assumes the discretization. If you don't, you still need the ordering.

bishtgautam:
@jeff-cohere Let's start with writing a binary restart file and for this, we can borrow PFLOTRAN's checkpoint approach that uses [`PetscBag`](https://petsc.org/release/manualpages/Sys/PetscBag/) as:

- A [`stepper_base_header_type` ](https://bitbucket.org/pflotran/pflotran/src/f0fe931c72c03580e489724afeb8c5451406b942/src/pflotran/timestepper_base.F90#lines-91) is registered in a `PetscBag` ([timestepper_base.F90#lines-934](https://bitbucket.org/pflotran/pflotran/src/f0fe931c72c03580e489724afeb8c5451406b942/src/pflotran/timestepper_base.F90#lines-934))
- Later on, the Vec (`flow_xx`) is dumped out in the `Viewer` [checkpoint.F90#lines-365](https://bitbucket.org/pflotran/pflotran/src/f0fe931c72c03580e489724afeb8c5451406b942/src/pflotran/checkpoint.F90#lines-365)

We can open a new issue specifically focussed on the binary restart format.

jeff-cohere:
I'm getting close to this and binary restart capability in [a feature branch](). In particular, binary checkpoints/restarts seem to be doing what they're supposed to. But when I try to write an HDF5 checkpoint file instead of a binary one, I see the following failure:

```
INFO: Writing checkpoint file checkpoints/ex2b-0000.h5...
[0]PETSC ERROR: --------------------- Error Message --------------------------------------------------------------
[...]

PR #134, opened 2024-02-21
Implemented checkpoints and restarts (binary working, HDF5 not)

### LINKED ISSUE #102: Add a simple binary restart mechanism
As a prelude to #41, we are adding a binary-format restart capability to RDycore using [PetscBag](https://petsc.org/release/manualpages/Sys/PetscBag/) and [PetscViewer](https://petsc.org/release/manualpages/Viewer/PetscViewer/).

@bishtgautam has pointed to examples in PFlotran to help us get started, including
* [a struct that stores simulation metadata](https://bitbucket.org/pflotran/pflotran/src/f0fe931c72c03580e489724afeb8c5451406b942/src/pflotran/timestepper_base.F90#lines-91) and [its registration in a PetscBag](https://bitbucket.org/pflotran/pflotran/src/f0fe931c72c03580e489724afeb8c5451406b942/src/pflotran/timestepper_base.F90#lines-934)
* [the dumping of solution data using a PetscViewer](https://bitbucket.org/pflotran/pflotran/src/f0fe931c72c03580e489724afeb8c5451406b942/src/pflotran/checkpoint.F90#lines-365)

## Design Questions

This list probably isn't complete.

1. What data and metadata are required for a complete restart of a simulation? Consider that we might want to allow restarts on different numbers of processes.
2. How does a user indicate that a simulation should be loaded from a restart instead of initialized? For example, do they simply add a `-restart` flag on the command line that indicates either a specific file or "the latest in a sequence"?
3. What exactly is the restart process? Do we go through the ordinary initialization process (loading the mesh, setting up Sections, etc) and then load solution and auxiliary data from the restart file?

jeff-cohere:
This PR introduces checkpoints and restarts to RDycore. We support both binary and HDF5 options. In its current state, this PR seems to demonstrate that binary checkpoints/restarts seem to be working well enough for us to recreate bit-for-bit simulation conditions, as reflected in new restart driver tests.

There's still an issue going on with the writing of HDF5 checkpoint files, which I've described in #41.

Closes #102


docs/user/input.md
```diff
@@ -377,9 +380,46 @@ so only the `flow` parameter is required. The presence of sediment concentration
...
+parameters in the `restart` section.
+
+* `format`: the format of the checkpoint files to be written. This can be either
+  `binary` (the default) or `hdf5`.
+* `interval`: the number of time steps taken between writing checkpoint files.
+  This must be a positive integer.
+* `prefix`: an optional prefix for checkpoint files. If omitted, the prefix for
+  checkpoint files is the prefix of the YAML input file name for the simulation.
+
+The name of a checkpoint file written at time step `N` is `<prefix>-N.<format>`,
```
bishtgautam, line 402: In E3SM, we have checkpoint/restart files named `<prefix>.<component>.r.YYYY-MM-DD-HHMMSS.nc`. So, in this PR we should change that filename to be `<prefix>.rdycore.r.N.<format>`. Later, we should allow an external driver (aka E3SM) to dictate the suffix of the restart file so that it would be something like `<prefix>.rdycore.r.<suffix>.<format>`.

driver/tests/swe_roe/ex2b_checkpoint.yaml
```diff
@@ -0,0 +1,88 @@
+# This test input corresponds to toy-problems/swe_roe/ex2b.c.
```
bishtgautam, line 1: How about updating the `ex2b.yaml` to include the `checkpoint` block instead of this new file?

bishtgautam (changes requested):
Can you add a test for the new restart capability?

jeff-cohere:
> Can you add a test for the new restart capability?

See [here](https://github.com/RDycore/RDycore/blob/jeff-cohere/restart/driver/tests/swe_roe/CMakeLists.txt#L46) for tests of the checkpointing and restarting process, with a call to `h5diff` to verify that the restarted simulation results match the original simulation results.

driver/tests/swe_roe/ex2b_checkpoint.yaml
jeff-cohere: Sure, I can consolidate the tests a bit.

bishtgautam (approved):

PR #135, opened 2024-02-29
Adds configuration files for PETSc

### LINKED ISSUE #127: Add configuration file

bishtgautam:
- Adds scripts for installing PETSc (32bit and 64bit) on the following DOE's machines:
   1. Perlmutter: CPU nodes
   2. Perlmutter: GPU nodes
   3. Frontier
   4. Aurora
 - Adds files that load modules corresponding to each PETSc installation
 - Adds a shell script (config/set_petsc_settings.sh) that does the following:
    1. Set PETSc environment variables (i.e. PETSC_DIR and PETSC_ARCH), and
    2. Loads the corresponding modules

The shell script can be run as:
1. For Perlmutter CPU or GPU nodes with support for 32bit (default) or 64bit integer
```
source config/set_petsc_settings.sh --pm <cpu|gpu> <--64bit>
```
2. For Frontier or Aurora with support for 32bit (default) or 64bit integer
```
source config/set_petsc_settings.sh <--64bit>
```

Fixes #127

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -110,3 +110,36 @@ The build directory has a structure that mirrors the source directory, and you
 can type `make` in any one of its subdirectories to do partial builds. In
 practice, though, it's safest to always build from the top of the build tree.
 
+
+## Preinstalled PETSc for RDycore on certain DOE machines
+
+The RDycore team supports installation of the model at following DOE machines:
+
+1. [Perlmutter](https://docs.nersc.gov/systems/perlmutter/) at NERSC
+2. [Frontier](https://docs.olcf.ornl.gov/systems/frontier_user_guide.html) at OLCF
+3. [Aurora](https://www.alcf.anl.gov/support-center/aurora/getting-started-aurora) at ALCF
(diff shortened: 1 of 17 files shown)


docs/installation.md
```diff
@@ -110,3 +110,36 @@ The build directory has a structure that mirrors the source directory, and you
 can type `make` in any one of its subdirectories to do partial builds. In
 practice, though, it's safest to always build from the top of the build tree.
 
+
+## Preinstalled PETSc for RDycore on certain DOE machines
+
```
jeff-cohere, line 115: This is great! Thanks for adding this section.

jeff-cohere: Do you think it's worth adding a README.md file to the `config/` directory briefly describing its contents, including the `install.*` scripts? Maybe it's not necessary, but it's nice to write this stuff down while it's in one's head.

bishtgautam: Good idea of adding a README.md. I will do that.

PR #136, opened 2024-03-01
Fixes HDF5 restart behavior

### LINKED ISSUE #41: Add restart capability

jeff-cohere:
See #41 for a summary of things that needed addressing. There are now quite a few tests, and at some point we may want to organize them a bit more intentionally, but that's a story for another time...

Closes #41

### DIFF
--- a/cmake/FindPETSc.cmake
+++ b/cmake/FindPETSc.cmake
@@ -22,7 +22,7 @@
 # For details see the accompanying COPYING-CMAKE-SCRIPTS file.
 #
 
-cmake_policy(VERSION 3.3)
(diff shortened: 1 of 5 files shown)


knepley (approved):
Looks good to me. I will get that Bag stuff done soon

bishtgautam (approved):
This is fantastic work. Thanks @jeff-cohere.

I completely agree that we need to reduce the total number of tests. I also found issues with some of the existing tests related to solution convergence. I will open a new issue regarding tests.

PR #139, opened 2024-03-04
Updates mesh creation to remove few assumptions

### LINKED ISSUE #138: Remove assumptions about number of vertices, edges, and neighbors of a cell
https://github.com/RDycore/RDycore/blob/main/src/rdymesh.c#L26-L28

```
  PetscInt vertices_per_cell  = 4;
  PetscInt edges_per_cell     = 4;
  PetscInt neighbors_per_cell = 4;
```

bishtgautam:
The max number of vertices/edges per cell and the max number of cells/edges per vertex are determined from the mesh instead of making an assumption about those values.

Fixes #138

### DIFF
--- a/src/rdymesh.c
+++ b/src/rdymesh.c
@@ -17,15 +17,15 @@ static PetscInt QUAD_ID_EXODUS = 5;
 
 /// Allocates and initializes an RDyCells struct.
 /// @param [in] num_cells Number of cells
+/// @param [in] nvertices_per_cell Maximum number of vertices per cell
+/// @param [in] nedges_per_cell Maximum number of edges per cell
 /// @param [out] cells A pointer to an RDyCells that stores allocated data.
 ///
 /// @return 0 on success, or a non-zero error code on failure
-PetscErrorCode RDyCellsCreate(PetscInt num_cells, RDyCells *cells) {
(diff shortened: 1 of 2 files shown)


src/rdymesh.c
```diff
@@ -1071,6 +1071,78 @@ static PetscErrorCode CreateCellConnectionVector(DM dm, RDyMesh *mesh) {
   PetscFunctionReturn(PETSC_SUCCESS);
 }
 
+/// Determines the max number of edges/vertices per cell and max number of
+/// cells/edges per vertices
+/// @param [in] dm A PETSc DM object
+/// @param [inout] mesh A pointer to an RDyMesh that is updated
+/// @return PETSC_SUCCESS on success
+PetscErrorCode DetermineMaxAttributesForCellsAndVertices(DM dm, RDyMesh *mesh) {
```
jeff-cohere, line 1079: This can also be marked `static` if it's not used anywhere else.

jeff-cohere (approved):
Looks great. One minor comment affecting function visibility.

src/rdymesh.c
bishtgautam: Good suggestion. I will update the PR.

PR #142, opened 2024-03-06
Ensembles, part 1: defining and running

### LINKED ISSUE #128: Add support for running ensemble calculations
Design points:

* We've decided to split the global MPI communicator into sub-communicators assigned to ensemble members within RDycore, allowing us to perform our own analysis
* We are adding an `ensemble` section to the YAML input file, [described here](https://github.com/RDycore/RDycore/blob/jeff-cohere/ensembles/docs/user/input.md#ensemble).
* We will allocate one GPU (device) to each ensemble member instead of sharing them between different ensemble members. The minimum ensemble size that uses an entire compute node is then determined by the number of available devices on that node, and larger ensembles should have a number of members that is a multiple of this minimum size.

Interested folks can follow the latest progress on this issue [in this branch](https://github.com/RDycore/RDycore/tree/jeff-cohere/ensembles).

jeff-cohere:
This PR adds an optional `ensemble` section to the input file (described in the User Guide, in the configuration file documentation) and allows you to run several ensemble members at once. No analysis capabilities are provided yet.

The `ensemble` section allows you to specify an ensemble size and list members of the ensemble explicitly, assigning parameters to be overridden in the same syntax as the parameters are specified in other sections. Each ensemble member is assigned the control parameters in the other sections and then the parameters of interest are overridden as specified in the `ensemble` section.

Ensembles can optionally be assigned a human-readable name--otherwise they get an autogenerated name like `ensemble_xyz`. Output for an ensemble is written to the `output/<ensemble-name>` directory, and checkpoints are written to `checkpoints/<ensemble-name>`. Of course, RDycore retains its original behavior if you don't have an `ensemble` section in your config file.

This feature still needs more testing, but it may be interesting to play with this stuff if you're interested. You can try this out by adding the following to the `ex2b.yaml` file:

```
ensemble:
  size: 5
  members:
  - materials:
    - name: smooth
      properties:
        manning:
          value: 0.15
  - materials:
    - name: smooth
      properties:
        manning:
          value: 0.20
  - materials:
    - name: smooth
      properties:
        manning:
          value: 0.25
  - materials:
    - name: smooth
      properties:
        manning:
          value: 0.30
  - materials:
    - name: smooth
      properties:
        manning:
          value: 0.35
```

### Bonus

Because of a low-level issue I encountered, I also added a `-pause` command line option that pauses RDycore when `RDySetup` is called, waiting for you to press Enter on rank 0. This command prints out all the PIDs for its global MPI ranks, which allows you to attach a debugger to take a closer look at what's happening if needed.

Closes #128


bishtgautam (approved):
@jeff-cohere This looks great. Let's add the sample ensemble yaml file in this PR, even though we may not run a test for the ensemble. Also, can you make sure that yaml file is copied to the build directory?

jeff-cohere:
Sounds good. I'll just make sure the ensembles are producing the right results, and fix them if they're not. :-)

bishtgautam:
Btw, what does the output look like when you add -log_view for the ensemble run? What about -log_view:filename?

jeff-cohere:
Sorry for the radio silence! I'm still chasing down some suble quirks of how things get overriden for ensemble members that are preventing answers from matching simulation results. But after that's sorted, I'll run with those arguments and post results.

Issue #146, opened 2024-03-07
Add cell centroid coordinates to the output


bishtgautam:
Add (x,y,z) corresponding to the cell centroid in the output. This will allow us to visualize the terrian elevation.

PR #147, opened 2024-03-07
Updates the XDMF output

bishtgautam:
The cell centroids are added to the H5 output files. This will enable the display
of elevation using XDMF.

Fixes https://github.com/RDycore/RDycore/issues/146

### DIFF
--- a/include/private/rdymeshimpl.h
+++ b/include/private/rdymeshimpl.h
@@ -195,7 +195,12 @@ typedef struct RDyMesh {
   // the maximum closure size for any cell (locally stored?)
   PetscInt maxClosureSize;
 
-  Vec coords_nat, cell_conn;
+  // for output: coordinates of vertices (in vertex natural order)
+  Vec coords_nat;
+  // for output: connections of vertices forming the cells (in cell natural order)
+  Vec cell_conn;
+  // for output: cell centroids (in cell natural order)
+  Vec xc, yc, zc;
 
 } RDyMesh;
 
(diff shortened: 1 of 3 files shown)


bishtgautam:
Now, we can visualize mesh elevation by displaying`ZC`.

zc

include/private/rdymeshimpl.h
```diff
@@ -195,7 +195,12 @@ typedef struct RDyMesh {
   // the maximum closure size for any cell (locally stored?)
   PetscInt maxClosureSize;
 
-  Vec coords_nat, cell_conn;
+  // for output: coordinates of vertices (in vertex natural order)
```
jeff-cohere, line 198: Do you think it would make sense to group these Vecs into an anonymous struct called `output` that makes clear wherever they're used that they're intended for output? For example:

```c
  struct {
    // ...
    Vec coords_nat;
    // ...
    Vec cell_conn;
    // ...
    Vec xc, yc, zc;
  } output;
```

Then you'd refer to these within an `RDyMesh` variable `mesh` as e.g. `mesh.output.coords_nat`, `mesh.output.xc`, and so on.

jeff-cohere: To be clear, this is just something to consider. I don't think it's strictly necessary.

jeff-cohere (approved):

include/private/rdymeshimpl.h
bishtgautam: It is a good idea to use a struct to consolidate the information. I will make the relevant change.

PR #150, opened 2024-03-12
Uses PetscCallCEED

### LINKED ISSUE #120: Define a macro called PetscCallCeed
We should define a macro `PetscCallCeed` based on [libCEED:examples/fluids/navierstokes.h#L27-L35](https://github.com/CEED/libCEED/blob/main/examples/fluids/navierstokes.h#L27-L35.) 

Then, convert RDycore core (e.g. https://github.com/RDycore/RDycore/blob/main/src/swe/swe_ceed.c#L274)

from
```
CeedCompositeOperatorCreate(ceed, flux_op);
```

to
```
PetscCallCeed(ceed, CeedCompositeOperatorCreate(ceed, flux_op));
```

based on [libCEED:examples/fluids/src/setuplibceed.c#L98](https://github.com/CEED/libCEED/blob/main/examples/fluids/src/setuplibceed.c#L98)

bishtgautam:
Checks the value returned by CEED using the Macro defined in PETSc.

Fixes #120

### DIFF
--- a/src/rdysetup.c
+++ b/src/rdysetup.c
@@ -931,7 +931,7 @@ PetscErrorCode RDySetup(RDy rdy) {
 
   // initialize CEED if needed
   if (rdy->ceed_resource[0]) {
-    CeedInit(rdy->ceed_resource, &rdy->ceed);
+    PetscCallCEED(CeedInit(rdy->ceed_resource, &rdy->ceed));
   }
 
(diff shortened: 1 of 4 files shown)


bishtgautam:
It turns out that PETSc already has a `PetscCallCEED()` macro. So, we don't have to define it in RDycore.

jeff-cohere (approved):

PR #153, opened 2024-03-13
Fixes error catching for setting PETSc on supported machines

### LINKED ISSUE #151: Improve error catching for config/set_petsc_settings.sh
@markadams reported: On Frontier, the following usage is not giving an error:

```
config/set_petsc_settings.sh --pm cpu
```

bishtgautam:
- Report an error when `--pm <cpu|gpu>` is specified on Perlmutter and Aurora.
- Additionally, adds support for the debugging version of PETSc for:
  - Perlmutter: CPU/GPU nodes and 32/64 bit
  - Frontier: Only for 32bit

Fixes #151

### DIFF
--- a/config/install.frontier.gnugpu.debug.32bit.sh
+++ b/config/install.frontier.gnugpu.debug.32bit.sh
@@ -1,0 +1,32 @@
+#!/bin/sh
+
+#source modules.pm-gpu.gnugpu
+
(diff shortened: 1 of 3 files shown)


jeff-cohere (approved):
Thanks for making it easier for us to run on these machines!

PR #159, opened 2024-03-22
 Switched to pkgconfig method for finding PETSc and friends.

jeff-cohere:
@bishtgautam , let's try this out for overcoming your current linking issues. If it works, we're off of the old CMake method of finding PETSc.


jeff-cohere:
@markadams , if/when you get a chance, can you make try this branch to see if works for you on the machines you're running on? This PR switches us to a new mechanism for finding PETSc on the machine, and I'd rather not mess up your environment if it doesn't work properly for you.

bishtgautam:
I tested out your branch on Perlmutter and the code compiled successfully for the two PETSc installations (i.e. CPU and GPU).

jeff-cohere:
Thanks, Gautam. That's probably good enough for us to merge this. I'll prioritize fixing whatever issues it causes.

PR #166, opened 2024-04-08
Uses PetscInt_FMT instead of `%d`

bishtgautam:
To avoid warning for PETSc configured with 64-bit integers, PetscInt_FMT is used.


jeff-cohere (approved):
👍

PR #169, opened 2024-04-15
Few minor updates to support E3SM-RDycore coupling for more than 1 proc

bishtgautam:
- Extracts the total or global number of cells in the mesh.
- Set natural cell ID to be the same as local cell ID when running on only one processor.

### DIFF
--- a/src/rdymesh.c
+++ b/src/rdymesh.c
@@ -818,13 +818,19 @@ static PetscErrorCode ComputeAdditionalCellAttributes(DM dm, RDyMesh *mesh) {
   PetscFunctionReturn(PETSC_SUCCESS);
 }
 
-static PetscErrorCode SaveNaturalCellIDs(DM dm, PetscInt num_cells, RDyCells *cells, PetscMPIInt rank) {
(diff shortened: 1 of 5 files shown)


src/rdymesh.c
```diff
@@ -878,6 +878,18 @@ static PetscErrorCode SaveNaturalCellIDs(DM dm, RDyCells *cells, PetscMPIInt ran
     PetscCall(VecDestroy(&natural));
     PetscCall(VecDestroy(&global));
     PetscCall(VecDestroy(&local));
+  } else {
+    PetscMPIInt commsize;
+    MPI_Comm    comm;
+    PetscCall(PetscObjectGetComm((PetscObject)dm, &comm));
+    PetscCallMPI(MPI_Comm_size(comm, &commsize));
```
jeff-cohere, line 885: You know, we could do the same thing to find the `rank` used in line 846, instead of passing it in.

jeff-cohere (approved):
Looks good! One very minor comment.

src/rdymesh.c
bishtgautam: Good point. Let me update the code accordingly.

PR #170, opened 2024-04-16
Fixed several compiler warnings related to PetscInt_FMT.

jeff-cohere:
I found these on my 64-bit integer build. :-)


bishtgautam (approved):
Ok, I just realized that `PetscInt_FMT` should be used only for `PetscInt` and not for other types of integer (e.g. `int`, `PetscMPIInt`)

PR #171, opened 2024-04-17
Exposing time units to interface header and adding to time-related functions

### LINKED ISSUE #160: Improve RDyGetTime()
`RDyGetTime(rdy, &time)` returns `time` in the units time is specified in the yaml file, but this is not very obvious. 

Here are two suggestions to improve:
- Replace `RDyGetTime` with `RDyGetTimeInSeconds()`, `RDyGetTimeInMinutes()`, etc, or
- Extend to accept another argument related to time units `RDyGetTime(rdy, 'seconds', &time)`

jeff-cohere:
This PR adds `RDyTimeUnit` to `rdycore.h` and adds a unit parameter to `RDyGetTime`, `RDyGetTimestep`, and `RDy{Get,Set}CouplingInterval`. Time quantities are still stored in config units with the configuration schema, and internal time is still measured in seconds.

Closes #160

### DIFF
--- a/driver/main.F90
+++ b/driver/main.F90
@@ -86,26 +86,27 @@ program rdycore_f90
 
   implicit none
 
-  character(len=1024) :: config_file
-  type(RDy)           :: rdy_
-  PetscErrorCode      :: ierr
-  PetscInt            :: n, step, iedge
-  PetscInt            :: nbconds, ibcond, num_edges, bcond_type
-  PetscReal, pointer  :: h(:), hu(:), hv(:), rain(:), bc_values(:), values(:), values_bnd(:)
-  PetscInt, pointer   :: nat_id(:), nat_id_bnd_cell(:)
-  PetscReal           :: time, time_step, prev_time, coupling_interval, cur_time
+  character(len=1024)  :: config_file
+  type(RDy)            :: rdy_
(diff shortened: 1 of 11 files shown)


bishtgautam:
If we want, can we in the driver code do the following?
```
 if (time_unit == TIME_SECONDS){
   ....
}
```

jeff-cohere:
> If we want, can we in the driver code do the following?
> 
> ```
>  if (time_unit == TIME_SECONDS){
>    ....
> }
> ```

Yes, that would work.

bishtgautam:
In that case, should we add a prefix `RDY_TIME_SECONDS`?

jeff-cohere:
> In that case, should we add a prefix `RDY_TIME_SECONDS`?

Oh, yeah, that's probably a good idea.

bishtgautam:
One more thought, should we add a helper function `RDyConvertTimeUnits(time, time_units_from, time_units_to`) like [`ConvertTimeToSeconds()`](https://github.com/RDycore/RDycore/blob/a272a5cc69d8c67be931c1c423b5e808d93dbf63/src/rdysetup.c#L36)?

bishtgautam (approved):
This looks great.

Issue #104, opened 2023-10-25
Add a test driver that supports the Method of Manufactured Solutions (MMS)


jeff-cohere:
To date, our CI testing doesn't actually verify any of our computed solutions. It would be nice if we could verify that RDycore's computed solutions converge at the expected rates, since the convergence rate is one of the most effective measures of correctness. It would certainly make me more confident about code changes!

@bishtgautam and I briefly discussed the possibility of putting together a test driver that accepts a special input format designed to accommodate MMS tests. Depending on the complexity of the desired tests, we could make it very easy to set up these tests by using tools that parse math expressions from text (like [tinyexpr](https://github.com/codeplea/tinyexpr)).

I don't know enough about what the science team has been investigating with MMS to draft a proposal. Gautam, if you know of materials and can link me to them, I could start thinking about this.

## TODO
- [x] Clean up some `RDySetup` code to make it easier for the main library and the MMS driver to share some things
- [x] Implement `mms` section parsing
- [x] Create `RDySetupMMS` as an alternative to `RDySetup` (in `mms.c`)
  - [x] analytic initial conditions
  - [x] analytic boundary conditions
  - [x] analytic sources
  - [x] analytic material properties (including those in source terms)
- [x] Update the `mms` driver to use the new system
- [x] provide a mechanism for convergence testing

jeff-cohere:
I did find [an ex2b_MMS.c](https://github.com/RDycore/toy-problems/blob/dongyu/MMS/swe/ex2b_MMS.c) file in the `toy-problems` repo.

knepley:
@jeff-cohere I recommend `muparser` for getting the functions. I have used it before, PETSc can build it automatically, and it is used in other projects.

bishtgautam:
@fdongyu has been undertaking model verification via:
1. MMS, and
2. Comparision against analytical solutions of SWE for multiple problems from https://hal.science/hal-00628246/file/SW_analytic-complements.pdf

Some information about the timeline. Over the next month, we will have multiple verification problems (~7) completed in the `toy-problems` repo. Then, we will start migrating those verification problems into RDycore repo and at that stage, we will be ready to start developing a verification driver.

bishtgautam:
Here is the link to our internal confluence page created by @fdongyu summarizing the validation problems for RDycore https://rdycore.atlassian.net/wiki/spaces/PD/pages/85458945/Flow+RDycore+validation+problems

jeff-cohere:
[...]

PR #172, opened 2024-04-19
Method of Manufactured Solutions, part 1: more expressive driver input

jeff-cohere:
This PR is the first of probably 2 in pursuit of #104, which we need to efficiently and comprehensively perform convergence testing.

This PR introduces a new MMS input format in which you can supply a text description of a specific manufactured solution. This text, which contains mathematical expressions for the solution and its derivatives, is parsed and compiled into a function that can be called in C.

A number of methods have been added to the C and Fortran interfaces to support this dynamically-configurable MMS setup. Accordingly, the C and Fortran drivers have both been modified to use this new input spec for MMS test problems. This approach gets rid of quite a lot of code, as you can see.

The next PR will incorporate convergence testing directly into the MMS driver programs.

## Disruptive change to input format

I had to make a slight change to our input format: **instead of specifying momentum as a 2-item array in flow initial/boundary conditions, we break momentum up into separate `x_momentum` and `y_momentum` fields. This makes it easier for us to parse math expressions for each of the momentum components.

## Error norms (updated)

On my machine,  the original C MMS driver gives the following output for the MMS test problem:

```
Avg-cell-area    : 0.2500001492054176
Avg-length-scale : 0.5000001492053953
Error-Norm-1     : 0.0099962646736428 0.0015251880380016 0.0015251880380016 
Error-Norm-2     : 0.0024488131544836 0.0004458908617737 0.0004458908617737 
Error-Norm-Max   : 0.0009621116663311 0.0002056314730604 0.0002056314730604
```

The new MMS driver gives this:

```
Avg-cell-area    : 0.2500001492054176
Avg-length-scale : 0.5000001492053953
Error-Norm-1     : 0.0076194713803673 0.0018090328041652 0.0018090328041652 
Error-Norm-2     : 0.0020462010292998 0.0005251756062723 0.0005251756062723 
Error-Norm-Max   : 0.0001749920885262 0.0002502143927563 0.0002502143927563
```

The updated MMS driver compares pretty well with the old one, but I won't really be happy till we measure some convergence rates.


bishtgautam:
@jeff-cohere I found two issues:

- **Issue-1**: The source term for the MMS is zero in this PR. I tested it via the following code change in this branch and the main branch:

```
diff --git a/src/swe/swe_petsc.c b/src/swe/swe_petsc.c
index 59cf08a..b29a2bf 100644
--- a/src/swe/swe_petsc.c
+++ b/src/swe/swe_petsc.c
@@ -601,6 +601,7 @@ PetscErrorCode AddSWESourceTerm(RDy rdy, Vec F) {
       }

       f_ptr[idx * ndof + 0] += swe_src_ptr[idx * ndof + 0];
+      printf("icell = %d; %e\n",icell,swe_src_ptr[idx * ndof + 0]);
       f_ptr[idx * ndof + 1] += -bedx - tbx + swe_src_ptr[idx * ndof + 1];
       f_ptr[idx * ndof + 2] += -bedy - tby + swe_src_ptr[idx * ndof + 2];
     }
```
I also changed the `mms_dx1.yaml` to `final_time       : 0.01` to reproduce the error in a single step.

- **Issue-2**: The `Problem1_DirichletValue()` in the `main` branch is wrong because the `(xc, yc)` being used for computing the Dirichlet BC is incorrect. Instead of the code at [driver/mms.c#L175-L178](https://github.com/RDycore/RDycore/blob/main/driver/mms.c#L175-L178), the `xc` and `yc` should be computed as:

```
    PetscReal xc = xc_edge[iedge];
    PetscReal yc = yc_edge[iedge];
```

bishtgautam:
> One explanation for this difference is that the new MMS driver does not use analytical expressions for the source terms--these are computed as is by the source operator.

I just read the above message after pasting my comment above, and I need to read the code to better understand the source operator.

bishtgautam:
Based on `PetscCall(RDyMMSEnforceBoundaryConditions(rdy, cur_time));` at [mms.c#L43](https://github.com/RDycore/RDycore/pull/172/files#diff-8d0a49b0c72189b3b4e9cc2e59e6c0f09bb92c1066a4a19b7311682e3627504fR43), how about a new function `PetscCall(RDyMMSEnforceSourceTerm(rdy, cur_time));`?

jeff-cohere:
I think we will need an MMS-specific source term for some problems, but for this particular one I think the same source term is used as we have implemented in the operator. Is this correct? If so, maybe we can try one of the other MMS problems with a different specified source term and implement the `RDyMMSEnforceSourceTerm` function you mention for it. Or would you rather just use the specified MMS source term even when it's the same as the discretized one?

jeff-cohere:
I just read your first comment. I had a feeling there are still issues--clearly something is going on with the source operator that I don't understand. I'll take a look at this tomorrow and see if I can implement the source term as specified by MMS and get the discretized source term to work for this particular problem--I can't think of a reason why it shouldn't work, but I probably don't understand everything I need to yet.

src/mms.c
```diff
@@ -0,0 +1,312 @@
...
+      // TODO: salinity and sediment initial conditions go here.
+
+      PetscInt l = 0;
+      for (PetscInt c = 0; c < region.num_cells; ++c) {
+        PetscInt cell_id = region.cell_ids[c];
+        if (3 * cell_id < n_local) {  // skip ghost cells
+          x_ptr[3 * cell_id]     = h[l];
+          x_ptr[3 * cell_id + 1] = h[l] * u[l];
+          x_ptr[3 * cell_id + 2] = h[l] * v[l];
+        }
```
bishtgautam, line 170: `++l;` is missing here.

src/rdydata.c
```diff
@@ -64,18 +64,18 @@ PetscErrorCode RDyGetBoundaryConditionFlowType(RDy rdy, const PetscInt boundary_
   PetscFunctionReturn(PETSC_SUCCESS);
 }
 
-PetscErrorCode RDySetDirichletBoundaryValues(RDy rdy, const PetscInt boundary_index, const PetscInt size, const PetscInt ndof,
-                                             PetscReal values[size]) {
+PetscErrorCode RDySetDirichletBoundaryValues(RDy rdy, const PetscInt boundary_index, const PetscInt num_edges, const PetscInt ndof,
```
bishtgautam, line 67: This is a really nice change.

src/rdymms.c
```diff
@@ -0,0 +1,398 @@
...
+
+  // We only need a single Dirichlet boundary condition, populated with
+  // manufactured solution data.
+  static RDyFlowCondition analytic_flow = {
+      .name = "analytic_bc",
+      .type = CONDITION_DIRICHLET,
+  };
+  analytic_flow.height     = rdy->config.mms.swe.solutions.h;
+  analytic_flow.x_momentum = rdy->config.mms.swe.solutions.u;  // NOTE: must multiply by h when enforcing!
+  analytic_flow.y_momentum = rdy->config.mms.swe.solutions.v;  // NOTE: must multiply by h when enforcing!
```
bishtgautam, line 30: I believe the comment `// NOTE: must multiply by h when enforcing!` is referring to the code, which is a bit later in this file:

```
      boundary_values[3 * e + 1] = h[e] * u[e];
      boundary_values[3 * e + 2] = h[e] * v[e];
```
If that is the case, why not simply use/define `analytic_flow.u` instead of `analytic_flow.x_momentum`?

bishtgautam (approved):
Great job, Jeff. This is a really cool PR. I have a minor comment. I will build PETSc on NERSC and Frontier with muparser tomorrow, and make sure we can build this branch on those machines.

src/rdymms.c
jeff-cohere: Excellent question. This is just how we originally defined boundary conditions (in terms of momentum instead of flow velocity). If we're willing to change how we specify boundary conditions to use flow velocities instead (which I think would be fine), we wouldn't have to do this tricky stuff. Let me know what you think about this.

In any case, I'd prefer to switch how we do boundary conditions in a separate PR.

PR #175, opened 2024-05-08
Mesh vertices are now snapped to z(x, y) even when unrefined.

jeff-cohere:
Gautam and I discussed this morning that the z coordinates of the mesh should be overwritten by the analytic bathymetry `z(x, y)` in all cases. This PR implements this change and some other small fixes.


bishtgautam (approved):
Thanks for this quick fix.

PR #176, opened 2024-05-10
Updates to support coupling with E3SM

bishtgautam:
Adds the following two functions to read data from a binary file and put the data in a local and global Vec:
- `RDyReadOneDOFLocalVecFromBinaryFile()`
- `RDyReadOneDOFGlobalVecFromBinaryFile()`

### DIFF
--- a/src/rdydata.c
+++ b/src/rdydata.c
@@ -417,33 +417,45 @@ PetscErrorCode RDyCreatePrognosticVec(RDy rdy, Vec *prog_vec) {
   PetscFunctionReturn(PETSC_SUCCESS);
 }
 
-// reads data for a single DOF from a binary file into a local Vec
-PetscErrorCode RDyReadOneDOFLocalVecFromBinaryFile(RDy rdy, const char filename[], Vec *local) {
+// reads data for a single DOF from a binary file into a global Vec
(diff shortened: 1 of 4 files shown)


src/rdydata.c
```diff
@@ -416,3 +416,47 @@ PetscErrorCode RDyCreatePrognosticVec(RDy rdy, Vec *prog_vec) {
   PetscCall(VecDuplicate(rdy->X, prog_vec));
   PetscFunctionReturn(PETSC_SUCCESS);
 }
+
+// reads data for a single DOF from a binary file into a global Vec
+PetscErrorCode RDyReadOneDOFGlobalVecFromBinaryFile(RDy rdy, const char filename[], Vec *global) {
```
jeff-cohere, line 421: What about `RDyRead{Global,Local}DOFFromBinaryFile` or `RDyRead{Global,Local}ComponentFromBinaryFile`? Would we ever want to read one component from a file storing a multi-component `Vec`?

jeff-cohere (approved):
Looks good. I just asked a question about the name.

src/rdydata.c
bishtgautam: 1. By including "OneDOF" in the function name, I was trying to indicate that we need to read the data using `rdy->aux_dm` instead of `rdy->dm`. However, I'm open to a better name.
2. I don't expect we would read only one component from a file storing a multi-component `Vec`?

PR #177, opened 2024-05-24
Removes one VecCopy and allocates another Vec at model initialization

bishtgautam:
1. Previously, a copy of the `U` Vector was made that was sent by PETSc's TS to the RHSFunction. This copy of the `U` Vec (`rdy->Soln`) was then sent to the source-sink CeedOperator. However, we must do a `GlobalToLocal` of `U` to get `U_local`. The `U_local` Vec is needed for the CeedOperator associated with internal and boundary edges. The copy of `U` to `rdy->Soln` is now avoided by sending `U_local` (instead of `rdy->Soln`) to the source-sink CeedOperator  (476c7ac916486e3aec0998b16edff34224733c8f).
2. The unused `rdy->Soln` is now deleted.
3. Detailed comments describing how CeedOperators are applied are added.
4. Renames `num_cells_local` to `num_onwned_cells`.
5. The Vec `F_dup` is now allocated during model initialization.

### DIFF
--- a/include/private/rdycoreimpl.h
+++ b/include/private/rdycoreimpl.h
@@ -164,6 +164,9 @@ struct _p_RDy {
   // residual vector
   Vec R;
 
+  // vector to store fluxes through internal and boundary edges
+  Vec F_dup;
+
   // source-sink vector
   Vec swe_src;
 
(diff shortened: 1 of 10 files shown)


jeff-cohere (approved):
Looks great!

bishtgautam:
@jedbrown I'm mostly seeking your review of https://github.com/RDycore/RDycore/commit/476c7ac916486e3aec0998b16edff34224733c8f. thanks.

PR #180, opened 2024-05-31
Updates the constant rainfall rate

bishtgautam:
- The default value of the constant rainfall rate is now changed to 0.0 from 1.0 mm/hr.
- A new command line argument is now added to specify the rainfall rate via `-constant_rain_rate <value>`.


bishtgautam:
The comparison for benchmark dam break problems fails without this code modification because a non-zero rainfall rate was automatically added to the problem.

jeff-cohere (approved):

PR #181, opened 2024-05-31
Fixes the global edge ID for serial runs

bishtgautam:
The global edge ID is correctly set for serial runs.

### DIFF
--- a/src/rdymesh.c
+++ b/src/rdymesh.c
@@ -438,8 +438,10 @@ static PetscErrorCode RDyEdgesCreate(PetscInt num_edges, RDyEdges *edges) {
 static PetscErrorCode RDyEdgesCreateFromDM(DM dm, RDyEdges *edges) {
   PetscFunctionBegin;
 
-  MPI_Comm comm;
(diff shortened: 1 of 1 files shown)


src/rdymesh.c
```diff
@@ -491,9 +493,12 @@ static PetscErrorCode RDyEdgesCreateFromDM(DM dm, RDyEdges *edges) {
   }
 
   // fetch global edge IDs.
-  ISLocalToGlobalMapping map;
-  PetscCall(DMGetLocalToGlobalMapping(dm, &map));
```
jeff-cohere, line 495: So we don't need these `*LocalToGlobal*` calls in the parallel case? Sorry--haven't looked at this stuff in a while. :-)

bishtgautam: This `*LocalToGlobal*` is only correct cells and not valid for edges because our DOFs are on cells. We need some new code (either in RDycore or in PETSc) to correctly set global edge IDs for parallel runs, which will be fixed in a future PR.

jeff-cohere (approved):

PR #189, opened 2024-06-16
Adds notes about the mesh

bishtgautam:
Adds notes about mesh formats (Exodus II and DMPlex's HDF5 v3.0.0) and
how the mesh is decomposed across ranks.


docs/developer/index.md
```diff
@@ -32,3 +32,4 @@ respectively) are all that is needed.
 * [Code Structure and Organization](organization.md)
 * [Developer Tools](tools.md)
 * [Integrating RDycore with E3SM](e3sm.md)
+* [Description About Mesh](mesh.md)
```
jeff-cohere, line 35: Mesh Description? Description of Mesh (or perhaps Description of Unstructured Grid)?

docs/developer/mesh.md
```diff
@@ -0,0 +1,270 @@
+# Mesh
+
+Here we describe the two mesh formarts that are used in RDycore. The two mesh
```
jeff-cohere, line 3: formarts -> formats

docs/developer/mesh.md
```diff
@@ -0,0 +1,270 @@
...
+formats that we use are:
+
+1. Exodus II
+2. DMPlex HDF5 v3.0.0
```
jeff-cohere, line 7: PETSc's DMPlex-specific, HDF5-based mesh format (version 3.0.0)

docs/developer/mesh.md
```diff
@@ -0,0 +1,270 @@
...
+
+RDycore has been tested for meshes that include triangular, quadrilateral, or
+both triangular and quadrilateral cells. The vertices of cells have coordinates
```
jeff-cohere, line 10: How about: The cell vertices have three-dimensional coordinates that incorporate topographic information.

docs/developer/mesh.md
```diff
@@ -0,0 +1,270 @@
...
+         / \
+        /   \
+      e5     e4
+     /   c1   \
+    /          \
+  v1 --- e3 --- v2
+```
+
+- The `elem_type` of the triangular cell, `c1`, is `TRI3`
+- It comprises of three vertices (i.e. `v1`, `v2`, `v3`)
```
jeff-cohere, line 32: "To comprise" is literally "to consist of", so this reads "It consists of of three vertices." Almost no one knows how to use the word "comprise" correctly, so I usually suggest avoiding it. :-)

For the first two bullets, how about

* The triangular cell `c1` consists of the three vertices `v1`, `v2`, and `v3`.
* The `elem_type` of `c1` is `TRI3`.

docs/developer/mesh.md
```diff
@@ -0,0 +1,270 @@
...
+- It has five edges sets. The first two edge sets form a plane and the remaining
+  three edge sets are lines as follows:
+  - `e1`: Comprises of the plane formed by `v1`, `v2`, `v3` (not shown above)
+  - `e2`: Comprises of the plane formed by `v1`, `v3`, `v2` (not shown above)
```
jeff-cohere, line 36: I think you are saying that `c1` is associated with five edges (edge sets are actually sets containing multiple edges in general). If that is true, I'm having trouble understanding what the edges `e1` and `e2` are. If I interpret your "plane formed by `v1`, `v2`, `v3` as the "(pseudo)vector formed by the cross product of `v1`, `v2`, `v3`", `e1` points "out of the screen", and correspondingly `e2` points "into the screen". But you're talking about planes and I'm talking about these pseudovectors. Is there a physical explanation of what the edges `e1` and `e2` represent?

docs/developer/mesh.md
```diff
@@ -0,0 +1,270 @@
...
+  - `e3`: Line from `v1` to `v2`
+  - `e4`: Line from `v2` to `v3`
+  - `e5`: Line from `v3` to `v1`
```
jeff-cohere, line 39: If edges are directed (i.e. if edge `e3` starts at `v1` and ends at `v2`), maybe we should call them "directed line segments" instead of "lines", which aren't directed and technical extend infinitely in either direction.

docs/developer/mesh.md
```diff
@@ -0,0 +1,270 @@
...
+- The `elem_type` of the quadrilateral cell, `c1`, is `SHELL4`
+- It comprises of four vertices (i.e. `v1`, `v2`, `v3`, `v4`)
+- It has six edges sets. The first two edge sets form a plane and the remaining
+  four edge sets are lines as follows:
+  - `e1`: Comprises of the plane formed by `v1`, `v2`, `v3`, `v4` (not shown above)
+  - `e2`: Comprises of the plane formed by `v1`, `v4`, `v3`, `v2` (not shown above)
+  - `e3`: Line from `v1` to `v2`
+  - `e4`: Line from `v2` to `v3`
+  - `e5`: Line from `v3` to `v4`
+  - `e6`: Line from `v4` to `v1`
```
jeff-cohere, line 64: The comments for the triangular cell apply here as well.

docs/developer/mesh.md
```diff
@@ -0,0 +1,270 @@
...
+
+### Example mesh with mixed element types
+
+![image](figures/quad_tri_mesh_exodus_ii.png)
+
+Above is an example of mesh that comprises of:
```
jeff-cohere, line 70: "an example of a mesh composed of" would work here.

docs/developer/mesh.md
```diff
@@ -0,0 +1,270 @@
...
+2. DMPlex HDF5 v3.0.0
+
+RDycore has been tested for meshes that include triangular, quadrilateral, or
+both triangular and quadrilateral cells. The vertices of cells have coordinates
+in 3D.
+
+## Exodus II Mesh Format
+
+The Exodus II mesh format (`.exo`) uses 1-based indices.
+
```
jeff-cohere, line 16: I think it's worth explaining here that the mesh consists of cells (triangular or quadrilateral), edges bounding the cells, and vertices at the edge's endpoints. Then we can say that the mesh boundary is represented by one or more "edge sets", and that these "edge sets" identify the edges that make up different parts of the boundary.

docs/developer/mesh.md
```diff
@@ -0,0 +1,270 @@
...
+- 14 vertices
+- Five edge sidesets
+  1. Right boundary
+  2. Left boundary
+  3. Top boundary
+  4. Bottom boundary
+  5. River
+
+The `.exo` mesh file uses NetCDF. The cells corresponding to the two element types
+(i.e. quadrilaterals and triangles) must be saved as two different data fields
```
jeff-cohere, line 82: Cells of different types (e.g. quadrilaterals and triangles) must be saved...

docs/developer/mesh.md
```diff
@@ -0,0 +1,270 @@
...
+  3, 7, 13,
+  7, 6, 13,
+  6, 2, 13,
+  6, 7, 14,
+  7, 11, 14,
+  11, 10, 14,
+  10, 6, 14 ;  
+```
+
+Each sideset is defined via two data fields:
```
jeff-cohere, line 105: Maybe we should briefly explain sidesets and how they relate to cells (elements) and edges at the top of the NetCDF format section.

docs/developer/mesh.md
```diff
@@ -0,0 +1,270 @@
...
+
+```
+
+## Domain Decomposition
+
+Below is the decomposition of the example mesh with mixed element types across two MPI ranks.
+
+![image](figures/quad_tri_mesh_dmplex_numbering_N2.png)
+
+The SWE physics in RDycore is currently implmented in two versions:
```
jeff-cohere, line 264: I'm not sure we need to mention SWE specifically here--we're likely to duplicate this structure in all of our physics. We could skip the PETSc implementation going forward, but personally I think we'll want it as a simpler reference implementation.

jeff-cohere (changes requested):
This is great, and it's going to be very helpful! Sorry for all the detailed nitpicking. Let me know if you have questions or would like me to suggest text for anything.

docs/developer/mesh.md
bishtgautam: "_SWE_" has been removed and I agree that we will continue to keep the PETSc implementation as it is simpler to debug.

docs/developer/mesh.md
```diff
@@ -0,0 +1,270 @@
...
+
+### DMPlex mesh numbering for a serial run
+
+DMPlex uses 0-based numbering. The mesh elements are numbered in the following order:
+
+1. Cells
+2. Vertices
+3. Edges. 
+
+DMPlex gives the following ID:
```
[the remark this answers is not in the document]
bishtgautam: Updated as suggested.

docs/developer/mesh.md
bishtgautam: Updated as suggested.

docs/developer/mesh.md
bishtgautam: Updated as suggested.

docs/developer/mesh.md
bishtgautam: Updated.

docs/developer/mesh.md
bishtgautam: Now uses the term "_an oriented plane_"

docs/developer/mesh.md
bishtgautam: Now uses the term "_a directed line_"

docs/developer/mesh.md
bishtgautam: Updated as suggested.

docs/developer/index.md
bishtgautam: Updated as suggested.

docs/developer/mesh.md
bishtgautam: fixed.

jeff-cohere (approved):

PR #190, opened 2024-06-17
Adding PETSc library directory to RPATH.

jeff-cohere:
This addresses an issue encountered by @donghuix, in which `libceed.dylib` wasn't being found by the linker.


bishtgautam (approved):

donghuix (approved):
Thank you for the fix.

PR #192, opened 2024-06-18
Removed `bed_friction` parameter from YAML spec.

### LINKED ISSUE #191: Misleading information about bed friction in the screen output
@donghuix reported the following screen output, which is misleading because bed friction is always turned on.

```
DETAIL: Physics:
DETAIL:   Flow:
DETAIL:     Bed friction: disabled
DETAIL:   Sediment model: disabled
DETAIL:   Salinity model: disabled
```

@jeff-cohere, We had previously discussed that will support multiple bed friction models that will be added in the `.yaml` as

```
bed_friction: <wylie93/xia18>
```

At present, we only support the `wylie93` bed friction term. So, how about we change the screen output to the following?

```
DETAIL: Physics:
DETAIL:   Flow:
DETAIL:     Bed friction model: Wylie93
```

jeff-cohere:
This parameter wasn't actually being used, and we always want bed friction in any case.

Closes #191

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,4 +4,5 @@ build-*/
 *.code-workspace
 compile_commands.json
 .cache/
+site/
 *.cgns
--- a/docs/common/input.md
+++ b/docs/common/input.md
@@ -431,7 +431,6 @@ time series data (and excluding checkpoint data). Relevant parameters are
 physics:
   flow:
     mode: swe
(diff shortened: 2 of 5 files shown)


bishtgautam (approved):
Thanks for the extremely fast turnaround on it 🙏

PR #201, opened 2024-08-19
Updating PETSc Docker image to v3.21.4.

jeff-cohere:


jeff-cohere:
I initially tried moving us to the latest Ubuntu Linux release, but this encountered problems in reading and writing HDF5 files in parallel. Hopefully that problem gets sorted out soon! So we're back to Ubuntu 22.04 for now.

knepley:
@jeff-cohere I think I know what this is. In some Linux, you have to https://petsc.org/main/manualpages/Viewer/PetscViewerHDF5SetCollective/ to PETSC_TRUE when HDF5 is built with parallel support. We have not had luck getting HDF5 to fix this. Let me know if this works.

jeff-cohere:
Thanks for the tip, @knepley ! I'll give that a shot when I find a moment.

jeff-cohere:
Okay, I added the command to enable MPI-IO collective reads/writes, but it doesn't seem to have addressed the issue we encounter, so I've backed off to Ubuntu 22.04 for now.

bishtgautam (approved):
Thank you for updating to the latest PETSc release.

PR #204, opened 2024-08-27
Updates configure files for PETSc 3.21.4

bishtgautam:
- Updates PETSc installation on Frontier and Perlmutter (CPU and GPU) to use PETSc 3.21.4.
- For Frontier, adds `stdc++` in `RDYCORE_LIBRARIES`.


jeff-cohere (approved):
Looks good to me. If you're satisfied that these are all working on their respective machines, we're good to merge.

PR #205, opened 2024-09-02
Updates YAML schema parser to official release version

jeff-cohere:
Our schema-based YAML parser got an official release, so I am updating to use that version, and making necessary changes to accommodate some functionality that was removed.


bishtgautam (approved):
I had to do the following before getting the code to compile without an error.

```
cd external/libcyaml/
rm -rf src/cyaml_proj-stamp tmp
cd ../../
git submodule update --init
```

jeff-cohere:
That makes sense, especially if you didn't remove your entire build tree. And submodules are tricky because they assume that you do `git submodule update` yourself. Thanks!

PR #210, opened 2024-09-09
Adds adaptive time stepping based on CFL condition

bishtgautam:
- Adds a timestep adaptivity block in the `.yaml` as:

```yaml
time:
  final_time        : 0.005
  coupling_interval : 0.001
  unit              : hours
  adaptivity:
    enable                 : true     # false or true. The values below are only used if is enable=true
    target_courant_number  : 0.6      # a target courant number
    max_increase_factor    : 2        # At max, the dt can be doubled to meet the target Courant number
    initial_time_step      : 0.00001  # initial timestep
```

- The time step for `TSSolve` is increased or decreased to match the `target_courant_number`.
- The time step is changed only at the coupling interval and **not within** the `TSSolve`.
- New tests for time step adaptivity are added.

### DIFF
--- a/driver/tests/swe_roe/CMakeLists.txt
+++ b/driver/tests/swe_roe/CMakeLists.txt
@@ -90,7 +90,7 @@ foreach(np 1 2) # test on 1, 2 processes
   endforeach()
 endforeach()
 
-# ex2b with ICs specified by a file
+# ex2b with ICs specified by a file and adaptive time step
 file(COPY
      ${CMAKE_CURRENT_SOURCE_DIR}/ex2b_ic_file_with_adapt.yaml
      ${MESH_DIR}/DamBreak_grid5x10.exo
--- a/driver/tests/swe_roe/CMakeLists.txt
+++ b/driver/tests/swe_roe/CMakeLists.txt
@@ -90,6 +90,21 @@ foreach(np 1 2) # test on 1, 2 processes
   endforeach()
 endforeach()
 
+# ex2b with ICs specified by a file
+file(COPY
+     ${CMAKE_CURRENT_SOURCE_DIR}/ex2b_ic_file_with_adapt.yaml
+     ${MESH_DIR}/DamBreak_grid5x10.exo
+     ${CONDITION_DIR}/DamBreak_grid5x10_wetdownstream.ic.${PETSC_ID_TYPE}.bin
+     ${MATERIALS_DIR}/manning_grid5x10.${PETSC_ID_TYPE}.bin
+     DESTINATION ${CMAKE_CURRENT_BINARY_DIR})
+foreach(np 1 2) # test on 1, 2 processes
+  foreach(config basic ceed preload)
+    foreach(lang c f90)
+      add_test(swe_roe_ex2b_ic_file_with_adapt_${lang}_np_${np}_${config} ${MPIEXEC} ${MPIEXEC_FLAGS} -n ${np} ${rdycore_${lang}_driver} ex2b_ic_file_with_adapt.yaml ${${config}_args})
+    endforeach()
+  endforeach()
+endforeach()
+
 # ex2b ensemble test -- (file copy only for now)
 file(COPY ${CMAKE_CURRENT_SOURCE_DIR}/ex2b-ensemble.yaml
      DESTINATION ${CMAKE_CURRENT_BINARY_DIR})
--- a/driver/tests/swe_roe/ex2b_ic_file_with_adapt.yaml
+++ b/driver/tests/swe_roe/ex2b_ic_file_with_adapt.yaml
@@ -17,7 +17,7 @@ time:
   final_time        : 0.005
   coupling_interval : 0.001
   unit              : hours
-  adaptivity:
+  adaptive:
     enable                 : true     # false or true. The values below are only used if is enable=true
     target_courant_number  : 0.6      # CFL
     max_increase_factor    : 2        # At max, the dt can be doubled while still allowing CFL be less than 0.8
--- a/driver/tests/swe_roe/ex2b_ic_file_with_adapt.yaml
+++ b/driver/tests/swe_roe/ex2b_ic_file_with_adapt.yaml
@@ -1,0 +1,60 @@
+# This test input corresponds to toy-problems/swe_roe/ex2b.c, but uses a file
+# to specify the initial conditions (and a corresponding Exodus mesh).
+
+physics:
+  flow:
+    mode: swe
+
+numerics:
+  spatial: fv
+  temporal: euler
+  riemann: roe
+
+logging:
+  level: debug
+
+time:
+  final_time        : 0.005
+  coupling_interval : 0.001
+  unit              : hours
+  max_step          : 240
+  adaptivity:
+    enable              : true  # false or true. The values below are only used if is enable=true
+    max_courant_number  : 0.6   #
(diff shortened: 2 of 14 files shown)


driver/tests/swe_roe/CMakeLists.txt
```diff
@@ -90,6 +90,21 @@ foreach(np 1 2) # test on 1, 2 processes
   endforeach()
 endforeach()
 
+# ex2b with ICs specified by a file
```
jeff-cohere, line 93: maybe something like

```
# ex2b with file-driven ICs and adaptive timestepping
```

driver/tests/swe_roe/ex2b_ic_file_with_adapt.yaml
```diff
@@ -0,0 +1,60 @@
...
+  riemann: roe
+
+logging:
+  level: debug
+
+time:
+  final_time        : 0.005
+  coupling_interval : 0.001
+  unit              : hours
+  adaptivity:
```
jeff-cohere, line 20: What do you think about using `adaptive` instead of `adaptivity` here? It's easier to spell, two-syllables shorter, and has essentially the same meaning. It's technically an adjective instead of a noun, but I think there's a limit to the value of consistency when it comes to practical considerations. :-)

src/swe/physics_swe.c
```diff
@@ -51,14 +51,15 @@ static PetscErrorCode InitMPITypesAndOps(void) {
...
+    const int      num_blocks             = 4;
+    const int      block_lengths[4]       = {1, 1, 1, 1};
+    const MPI_Aint block_displacements[4] = {
         offsetof(CourantNumberDiagnostics, max_courant_num),
         offsetof(CourantNumberDiagnostics, global_edge_id),
         offsetof(CourantNumberDiagnostics, global_cell_id),
+        offsetof(CourantNumberDiagnostics, is_set),
     };
-    MPI_Datatype block_types[3] = {MPI_DOUBLE, MPI_INT, MPI_INT};
+    MPI_Datatype block_types[4] = {MPI_DOUBLE, MPI_INT, MPI_INT, MPI_C_BOOL};
```
jeff-cohere, line 62: Technically, this field is a `PetscBool`, which is an enum type, and not a C `bool` type, but I think both can be demoted to `int` as needed, so this shouldn't be an issue.

src/rdyadvance.c
```diff
@@ -244,14 +245,69 @@ PetscErrorCode RDyAdvance(RDy rdy) {
...
+          RDyLogDebug(rdy, "Increasing dt from %f [%s] to %f [%s]", dt_old, units, dt_new, units);
+        }
+
+        // update the timestep
+        rdy->dt = dt;
+
+      } else {
+        // decrease the timestep
+        PetscReal factor = time_adap->target_courant_number / cnum_diags->max_courant_num;
+
```
jeff-cohere, line 285: You can get rid of lines 285-286 and just group both lines of code under one comment.

src/rdyadvance.c
```diff
@@ -284,7 +344,7 @@ PetscBool RDyFinished(RDy rdy) {
   PetscReal time_in_unit = ConvertTimeFromSeconds(time, rdy->config.time.unit);
   PetscInt  step;
   PetscCall(TSGetStepNumber(rdy->ts, &step));
-  if ((time_in_unit >= rdy->config.time.final_time) || (step >= rdy->config.time.max_step)) {
+  if ((time_in_unit >= rdy->config.time.final_time)) {  // || (step >= rdy->config.time.max_step)) {
```
jeff-cohere, line 347: Shouldn't we leave the `max_step` check in here for the fixed-timestep case? We can set `max_step` to `INT_MAX`, which is the maximum (32-bit) value for C integers. This should be appropriate for 64-bit integers, too, because I don't think we want to take 2e9 timesteps in any simulation(!)

src/swe/physics_swe.c
jedbrown: `sizeof(bool)` is 1, but `sizeof(PetscBool)` is 4 to match Fortran
```c
#define PetscBool  logical(kind=4)
```
You can use `MPIU_BOOL` to send `PetscBool`.

jeff-cohere: Thanks, Jed!

src/swe/physics_swe.c
```diff
@@ -367,16 +490,18 @@ PetscErrorCode RHSFunctionSWE(TS ts, PetscReal t, Vec X, Vec F, void *ctx) {
     }
   }
 
-  // write out debugging info for maximum courant number
+  // find the maximum courant number and write out the debugging info
   if (rdy->config.logging.level >= LOG_DEBUG) {
-    MPI_Allreduce(MPI_IN_PLACE, &courant_num_diags, 1, courant_num_diags_type, courant_num_diags_op, rdy->comm);
+    PetscCall(SWEFindMaxCourantNumber(rdy));
```
jeff-cohere, line 495: Should we check to see whether the Courant number is set here, to avoid an unnecessary recalculation?

src/rdyadvance.c
jeff-cohere: Aha, I see we have `max_step` set to `INVALID_INT` by default, so maybe this check should be done elsewhere (if it's not already).

src/swe/physics_swe.c
jedbrown: I see `CourantNumberDiagnostics` uses `PetscReal` and `PetscInt`.
```suggestion
    MPI_Datatype block_types[4] = {MPIU_REAL, MPIU_INT, MPIU_INT, MPIU_BOOL};
```

jeff-cohere (changes requested):
This looks good! @jedbrown pointed out some changes to apply, and I made some minor suggestions for your consideration.

src/rdyadvance.c
bishtgautam: I consolidated more common code outside the if-else loop.

src/swe/physics_swe.c
bishtgautam: I updated the comment to indicate that we must find the latest Courant number and disregard the previously computed value.

src/swe/physics_swe.c
```diff
@@ -51,14 +51,15 @@ static PetscErrorCode InitMPITypesAndOps(void) {
...
+    const int      num_blocks             = 4;
+    const int      block_lengths[4]       = {1, 1, 1, 1};
+    const MPI_Aint block_displacements[4] = {
         offsetof(CourantNumberDiagnostics, max_courant_num),
         offsetof(CourantNumberDiagnostics, global_edge_id),
         offsetof(CourantNumberDiagnostics, global_cell_id),
+        offsetof(CourantNumberDiagnostics, is_set),
     };
-    MPI_Datatype block_types[3] = {MPI_DOUBLE, MPI_INT, MPI_INT};
+    MPI_Datatype block_types[4] = {MPI_DOUBLE, MPI_INT, MPI_INT, MPIU_BOOL};
```
jeff-cohere, line 62: Jed suggested using `MPIU_REAL` insteadl of `MPI_DOUBLE` here. I think that handles both single and double precision (in case it ever matters).

jeff-cohere (changes requested):
Looks great! Thanks, Gautam. Just one more tiny fix (`MPI_DOUBLE` -> `MPIU_REAL`) and we can merge.

jeff-cohere (approved):

PR #212, opened 2024-09-12
submodules: use https to avoid clone failures when people aren't using ssh

jedbrown:
We don't need to push directly to these submodules and developers can always change the path for those submodules (e.g., by editing `.git/config`). It's pretty widely recommended (e.g., [github's blog](https://github.blog/open-source/git/working-with-submodules/)) to use https for submodules.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,9 +1,9 @@
 [submodule "external/cmocka"]
 	path = external/cmocka
-	url = git@github.com:clibs/cmocka.git
+	url = https://github.com/clibs/cmocka.git
 [submodule "external/libyaml"]
 	path = external/libyaml
-	url = git@github.com:yaml/libyaml.git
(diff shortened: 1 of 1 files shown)


jeff-cohere:
This is a good idea--I've been in the habit of using SSH because some laboratories have very arcane IT environments where HTTPS doesn't work(!).

jeff-cohere:
@bishtgautam , do you mind checking out this branch and seeing whether you have issues updating submodules? If you know that HTTPS submodules work for PNNL's network, no need to check here.

jedbrown:
If you're going to test, do it in a fresh clone, not an existing repository (`.git/config` holds the URL for submodules that have already been initialized; the content of `.gitmodules` isn't relevant unless you deinit those modules).

jeff-cohere (approved):

PR #213, opened 2024-09-12
Now building libcyaml in our own build system, for simplicity.

jeff-cohere:
For some reason, CMake wasn't properly configuring and building libcyaml as an external project, so I replaced that setup with a simple library build. It's a little disturbing to think that there may be regressions in CMake's ability to build non-CMake external projects, but this is the world we live in...


jeff-cohere:
@jedbrown , feel free to merge this if you're happy with it.

bishtgautam (approved):

jedbrown (approved):
Works fine for me.

Issue #214, opened 2024-09-13
Remove VLA allocations from code


jeff-cohere:
A recent [Slack conversation](https://rdycore.slack.com/archives/C035S99KZA5/p1726201331633899) unearthed the topic of Variable-Length Arrays and their use in PETSc. We concluded that VLA allocations should be removed (since they should really be confined to "small" sizes, in order not to exceed the size of the stack), and that this is hard to estimate in general for anything but the smallest size parameters (like `dim` or number of strided components).

However, we can keep VLA pointers (as expressed in function parameters), which C99 introduced as a handy feature for indexing multidimensional arrays.

PR #215, opened 2024-09-13
Adds a new struct for edge-related quantities

bishtgautam:
A new struct is introduced to save geometric, flux, and courant number
for internal and boundary edges instead of using variable length arrays.

### DIFF
--- a/include/private/rdycoreimpl.h
+++ b/include/private/rdycoreimpl.h
@@ -234,7 +234,7 @@ PETSC_INTERN PetscErrorCode ReadCheckpointFile(RDy, const char *);
 
 // time series
 PETSC_INTERN PetscErrorCode InitTimeSeries(RDy);
-PETSC_INTERN PetscErrorCode AccumulateBoundaryFluxes(RDy, RDyBoundary boundary, PetscInt size, PetscInt ndof, PetscReal[size][ndof]);
+PETSC_INTERN PetscErrorCode AccumulateBoundaryFluxes(RDy, RDyBoundary boundary, PetscInt size, PetscInt ndof, PetscReal[size * ndof]);
 PETSC_INTERN PetscErrorCode WriteTimeSeries(TS, PetscInt, PetscReal, Vec, void *);
 PETSC_INTERN PetscErrorCode DestroyTimeSeries(RDy);
 
--- a/include/private/rdycoreimpl.h
+++ b/include/private/rdycoreimpl.h
@@ -234,7 +234,7 @@ PETSC_INTERN PetscErrorCode ReadCheckpointFile(RDy, const char *);
 
 // time series
 PETSC_INTERN PetscErrorCode InitTimeSeries(RDy);
-PETSC_INTERN PetscErrorCode AccumulateBoundaryFluxes(RDy, RDyBoundary boundary, PetscInt size, PetscInt ndof, PetscReal[size * ndof]);
+PETSC_INTERN PetscErrorCode AccumulateBoundaryFluxes(RDy, RDyBoundary boundary, PetscInt size, PetscInt ndof, PetscReal *);
 PETSC_INTERN PetscErrorCode WriteTimeSeries(TS, PetscInt, PetscReal, Vec, void *);
 PETSC_INTERN PetscErrorCode DestroyTimeSeries(RDy);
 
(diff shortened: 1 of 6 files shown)


bishtgautam:
This PR partially addresses the issue mentioned in https://github.com/RDycore/RDycore/issues/214.

include/private/rdycoreimpl.h
```diff
@@ -234,7 +234,7 @@ PETSC_INTERN PetscErrorCode ReadCheckpointFile(RDy, const char *);
 
 // time series
 PETSC_INTERN PetscErrorCode InitTimeSeries(RDy);
-PETSC_INTERN PetscErrorCode AccumulateBoundaryFluxes(RDy, RDyBoundary boundary, PetscInt size, PetscInt ndof, PetscReal[size][ndof]);
+PETSC_INTERN PetscErrorCode AccumulateBoundaryFluxes(RDy, RDyBoundary boundary, PetscInt size, PetscInt ndof, PetscReal[size * ndof]);
```
jeff-cohere, line 237: Actually, I think we can keep the VLA pointers--they aren't related to stack allocations. Does this produce different behavior?

bishtgautam: If I don't make the change, we will get a compiler error because the variable that is being passed to this function is a 1D pointer at https://github.com/RDycore/RDycore/pull/215/files#diff-6cfdc6b20c73c1aa99a7cbfcfda14bd9b0a765a72fa9942cda460313483e277fR17

jedbrown: This hardly means anything in the declaration (in principle a compiler could check the invariant regarding allocation size, at least when `size` and `ndof` are compile-time constants, but I don't think I've seen that happen. I usually pass flat arrays through the public interface and dimension it with a VLA pointer within a function. If the build is happy (and we don't need to call the interface from C++), then either should be fine.

jedbrown: It's your choice where to cast `(PetscReal (*)[ndof])flux`, either inside the function or at the call site. In general, I've found passing higher-dimensioned arrays often ends up brittle for C reasons so I usually pass flat arrays through public interfaces.

bishtgautam: I will change it to
```
PETSC_INTERN PetscErrorCode AccumulateBoundaryFluxes(RDy, RDyBoundary boundary, PetscInt size, PetscInt ndof, PetscReal*);
```

src/swe/swe_petsc.c
```diff
@@ -10,30 +10,75 @@ static const PetscReal GRAVITY = 9.806;
 
 /// For computing fluxes, allocates structs to hold values left and right
 /// of internal and boundary edges. This must be called before CreatePetscSWESource.
-PetscErrorCode CreatePetscSWEFlux(PetscInt num_internal_edges, PetscInt num_boundaries, RDyBoundary boundaries[num_boundaries], void **petsc_rhs) {
+PetscErrorCode CreatePetscSWEFlux(RDyCells *cells, RDyEdges *edges, PetscInt ncomp, PetscInt num_internal_edges, PetscInt num_boundaries,
+                                  RDyBoundary boundaries[num_boundaries], PetscBool ceed_enabled, void **petsc_rhs) {
```
jeff-cohere, line 14: Things are getting pretty busy in these function signatures, but I think this can be addressed in the upcoming refactor/cleanup.

src/time_series.c
```diff
@@ -118,7 +118,7 @@ PetscErrorCode InitTimeSeries(RDy rdy) {
 
 // Accumulates boundary fluxes on the given boundary from the given array of
 // fluxes on boundary edges.
-PetscErrorCode AccumulateBoundaryFluxes(RDy rdy, RDyBoundary boundary, PetscInt size, PetscInt ndof, PetscReal fluxes[size][ndof]) {
+PetscErrorCode AccumulateBoundaryFluxes(RDy rdy, RDyBoundary boundary, PetscInt size, PetscInt ndof, PetscReal *fluxes) {
```
jeff-cohere, line 121: We could in fact "reshape" the array within this function using VLA pointers, which would make array access nicer. I'll do this in the refactor/cleanup.

include/private/rdysweimpl.h
```diff
@@ -11,11 +11,20 @@ typedef struct {
   PetscReal *u, *v;        // diagnostic variables
 } RiemannDataSWE;
 
+typedef struct {
+  PetscInt   N;        // number of data values
+  PetscReal *cn, *sn;  // cosine and sine value of a edge
```
jedbrown, line 16: Perhaps "x and y components of edge normal"?

jeff-cohere: We can scoop this up in the refactor.

jedbrown: I think that's also a bit ambiguously worded (is it the edge or the normal to the edge?). Not critical now, but if we move to a curvilinear coordinate system, we'll want to be more precise about what quantity we're storing.

PR #216, opened 2024-09-16
Allows model output at a fixed temporal frequency

### LINKED ISSUE #211: Allow output to be written out at a specified temporal frequency
Currently, output can be written out at a specified number of model steps. e.g. `interval: 100`
tells the model to write output every 100 model steps. This is okay for the case when the model
takes a constant time step. However, with time step adaptivity (https://github.com/RDycore/RDycore/pull/210),
we need the ability to specify output to be written at fixed time intervals (say 30 mins).

bishtgautam:
- Changes in the output block format:

   - The older output block only supports output to be written per interval step and is given as:
```
output:
  format: xdmf
  interval: 10000
```

   - The new output block supports output to be written 
      1. after a specified number of steps and/or 
      2. after a specified time frequency

```
output:
  format: xdmf
  step_interval: 10000
  time_interval: 600     # an integer value
  time_unit: seconds
```

- It is assumed that `time_interval` would be a multiple of `coupling_time` and the output requested **cannot** be within a `TSSolve`.

- If `coupling_time` is not specified, it is set as `time_interval`.

- Deletes a test based on ex2b and adds a more appropriate test corresponding to Houston Harvey flooding.

Fixes #211


jeff-cohere:
Question: If `time_interval` is given and `coupling_time` isn't, should it matter? It may be true that the two are related by a factor if both are set, but if we don't have a `coupling_time`, isn't it okay to have any `time_interval`?

docs/common/input.md
```diff
@@ -415,7 +415,9 @@ time series data (and excluding checkpoint data). Relevant parameters are
     * `binary`: output is written using PETSc's binary data format
     * `xdmf`: output is written to the [XDMF](https://xdmf.org/index.php/XDMF_Model_and_Format) format
     * `cgns`: output is written to the [CFD General Notation System (CGNS)](https://cgns.github.io/) format
-* `interval`: the number of time steps between output dumps. Default value: 0 (no output)
+* `step_interval`: the number of time steps between output dumps. Default value: 0 (no output)
+* `time_interval`: the temporal frequency between output dumps.Default value: 0 (no output)
+* `time_unit`: units of temporal frequency output.
```
jeff-cohere, line 420: is `time_unit` required when `time_interval` is given? Or does it have a default?

bishtgautam:
If `coupling_time` isn't specified, we set `coupling_time` as `final_time`. We can certainly set it to be `time_interval`.

docs/common/input.md
jeff-cohere: Also: should we be allowing/encouraging the use of multiple time units in single input files? I know PFLOTRAN allows this because its input format associates units with each quantity, but without such a feature this could get confusing.

bishtgautam: Good catch, as I forgot to add a check if `time_unit` is specified or not.

include/private/rdyconfigimpl.h
```diff
@@ -159,10 +159,14 @@ typedef struct {
...
-  PetscInt        batch_size;   // number of timesteps per output file (if available)
-  RDyTimeSeries   time_series;  // time series appended to text (.dat) files
+  PetscBool       enable;            // true if output is requested
+  RDyOutputFormat format;            // file format
+  PetscInt        step_interval;     // output interval [steps between outputs]
+  PetscReal       time_interval;     // temporal interval at which output is written
+  RDyTimeUnit     time_unit;         // unit of time for temporal interval
+  PetscInt        batch_size;        // number of timesteps per output file (if available)
+  RDyTimeSeries   time_series;       // time series appended to text (.dat) files
+  PetscReal       prev_output_time;  // previous time at which output was written
```
jeff-cohere, line 169: This isn't a config parameter, right? We might want to put this somewhere in the model instead, since it's state information.

src/yaml_input.c
```diff
@@ -873,12 +878,36 @@ static PetscErrorCode ValidateConfig(MPI_Comm comm, RDyConfig *config, PetscBool
...
+    PetscCheck((config->output.step_interval >= 0), comm, PETSC_ERR_USER, "Output step interval must be specified as a positive number of steps.");
+    PetscCheck((config->output.time_interval >= 0.0), comm, PETSC_ERR_USER, "Output time interval must be specified as a positive number of steps.");
+    PetscCheck((config->output.batch_size == 0) || (config->output.format != OUTPUT_BINARY), comm, PETSC_ERR_USER,
+               "Binary output does not support output batching");
+    if ((config->output.batch_size == 0) && (config->output.format != OUTPUT_NONE) && config->output.format != OUTPUT_BINARY) {
+      config->output.batch_size = 1;
+    }
+
+    if (config->output.time_interval) {
+      printf("config->time.coupling_interval = %f; coupling_interval_was_specified = %d\n", config->time.coupling_interval,
```
jeff-cohere, line 901: Leftover print statement here. If we don't need this, you can probably get rid of `coupling_interval_was_specified`.

jeff-cohere (approved):
Looks good. Storing the previous step at which output occurred in the config data structure mixes config parameters and state, but I don't think it's a big deal.

jeff-cohere:
Aha, looks like there's an issue with the Harvey test using 64-bit indices.

jeff-cohere (approved):

PR #217, opened 2024-09-17
Removing VLA allocations from code

### LINKED ISSUE #214: Remove VLA allocations from code

jeff-cohere:
This PR adds compiler flags that disallow C99 VLA code and gets rid of VLA declarations in function declarations, as well as all VLA allocations.

We can go back and add in VLA multidimensional array casting where needed with pragmas (see, for example, `swe_ceed.c`).

Closes #214

### DIFF
--- a/src/swe/swe_ceed.c
+++ b/src/swe/swe_ceed.c
@@ -3,6 +3,10 @@
 
 #include "swe_ceed_impl.h"
 
+// this value stores the maximum number of components needed (between solution
+// vector components, source vector components, geometric components, etc)
+#define MAX_NUM_COMPONENTS 4
+
 // frees a data context allocated using PETSc, returning a libCEED error code
 static int FreeContextPetsc(void *data) {
   if (PetscFree(data)) return CeedError(NULL, CEED_ERROR_ACCESS, "PetscFree failed");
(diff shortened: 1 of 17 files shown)


jeff-cohere:
> Is there a warning we could enable to check for this? (I'm not aware of any.) It's hard to grep for.

For VLA usage? Yes, I've enabled it and set it to be converted to a compile error. See the top-level `CMakeLists.txt`.

src/swe/swe_ceed.c
```diff
@@ -3,6 +3,14 @@
 
 #include "swe_ceed_impl.h"
 
+// CEED uses C99 VLA features for shaping multidimensional
+// arrays, which don't have the same drawbacks as VLA allocations.
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wvla"
+
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wvla"
```
bishtgautam, line 12: If one complies with an intel compiler, would we need to add a new pragma?

bishtgautam (approved):
Looks good to me and I have only one comment regarding intel compiler.

src/swe/swe_ceed.c
bishtgautam: Btw, we are presently using gcc on Perlmutter and Frontier.

bishtgautam: And this branch compiles just fine on both Perlmutter and Frontier.

jeff-cohere: These pragmas disable the warnings that we've put in place for GCC and Clang. I'm not sure if Intel even has a warning option for VLAs, but if we don't enable such a thing, we don't need the pragmas here.

GCC and Clang warnings should be enough to keep us from using VLAs unintentionally, because our CI system probably uses GCC in all cases.

PR #226, opened 2024-10-17
Operator refactor part 1: renaming and regrouping

jeff-cohere:
This PR moves solver-related data around, renaming things to be a bit more functionally clear.

The next step in this refactor process is to gather all the high-level solver-related logic into `solvers.c` so we can start making more complicated composite operators, such as those for the Hairsine-Rose equations that couple sediment transport to the shallow water equations.

### DIFF
--- a/include/private/rdycoreimpl.h
+++ b/include/private/rdycoreimpl.h
@@ -176,22 +176,20 @@ struct _p_RDy {
 
   // CEED (device) solver data
   struct {
-    // CEED context and resource name
-    Ceed context;
+    // CEED resource name -- used to determine the backend
     char resource[PETSC_MAX_PATH_LEN];
 
-    // CEED (composite) operators for fluxes and sources
+    Ceed context;
+
     CeedOperator flux_operator;
     CeedOperator source_operator;
 
-    // on-device vectors
     CeedVector u_local;
     CeedVector rhs, sources;
 
     CeedScalar dt;
 
-    // host vector storing fluxes
-    Vec host_fluxes;
+    Vec host_fluxes;  // edge fluxes copied to host for diagnostics
   } ceed;
 
   // PETSc (host) solver data
--- a/include/private/rdycoreimpl.h
+++ b/include/private/rdycoreimpl.h
@@ -171,8 +171,8 @@ struct _p_RDy {
   // host solution vectors (global and local)
   Vec u_global, u_local;
 
-  // host residual vector
-  Vec residual;
+  // host right-hand-side (residual) vector
+  Vec rhs;
 
   // CEED (device) solver data
   struct {
--- a/include/private/rdycoreimpl.h
+++ b/include/private/rdycoreimpl.h
@@ -168,17 +168,40 @@ struct _p_RDy {
   // time₋stepping solver
   TS ts;
 
-  // solution vectors (global and local)
-  Vec X, X_local;
+  // host solution vectors (global and local)
+  Vec u_global, u_local;
 
-  // residual vector
-  Vec R;
+  // host residual vector
+  Vec residual;
 
-  // vector to store fluxes through internal and boundary edges
-  Vec F_dup;
+  // CEED (device) solver data
+  struct {
+    // CEED context and resource name
+    Ceed context;
+    char resource[PETSC_MAX_PATH_LEN];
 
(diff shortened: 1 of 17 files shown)


include/private/rdycoreimpl.h
```diff
@@ -168,17 +168,38 @@ struct _p_RDy {
...
 
-  // solution vectors (global and local)
-  Vec X, X_local;
+  // host solution vectors (global and local)
+  Vec u_global, u_local;
 
-  // residual vector
-  Vec R;
+  // host residual vector
+  Vec residual;
```
bishtgautam, line 175: How about we rename `residual` to `rhs`? This way it would be consistent with the name used in `ceed` struct and  we are using this vector for `TSSetRHSFunction`.

src/rdysetup.c
```diff
@@ -474,13 +473,13 @@ PetscErrorCode InitBoundaries(RDy rdy) {
         RDySurfaceCompositionSpec surface_comp = rdy->config.surface_composition[isurf_comp];          \
         if (!strcmp(surface_comp.region, region.name)) {                                               \
           if (mat_prop_vec) {                                                                          \
-            PetscScalar *x_ptr;                                                                        \
-            PetscCall(VecGetArray(mat_prop_vec, &x_ptr));                                              \
+            PetscScalar *u_ptr;                                                                        \
+            PetscCall(VecGetArray(mat_prop_vec, &u_ptr));                                              \
```
bishtgautam, line 477: Since `mat_prop_vec` is a 1 DOF vector, how about renaming `u_ptr` to `u1d_ptr` (or something else) that makes it a bit explicit that this is a pointer to 1 DOF vector.

src/rdysetup.c
```diff
@@ -793,9 +792,9 @@ static PetscErrorCode InitSolution(RDy rdy) {
           for (PetscInt c = 0; c < region.num_cells; ++c) {
             PetscInt cell_id = region.cell_ids[c];
             if (ndof * cell_id < n_local) {  // skip ghost cells
-              x_ptr[3 * cell_id]     = mupEval(flow_ic.height);
-              x_ptr[3 * cell_id + 1] = mupEval(flow_ic.x_momentum);
-              x_ptr[3 * cell_id + 2] = mupEval(flow_ic.y_momentum);
+              u_ptr[3 * cell_id]     = mupEval(flow_ic.height);
+              u_ptr[3 * cell_id + 1] = mupEval(flow_ic.x_momentum);
+              u_ptr[3 * cell_id + 2] = mupEval(flow_ic.y_momentum);
```
bishtgautam, line 797: Since we are editing these lines, how about changing `u_ptr[3 * ...` to `u_ptr[ndof * ...`?

bishtgautam (approved):
These are great changes (👍)  and I have a few minor suggestions.

src/rdysetup.c
jeff-cohere: How about `prop_ptr`?

bishtgautam: That sounds good.

src/swe/physics_swe.c
```diff
@@ -269,25 +269,25 @@ static PetscErrorCode RDyCeedOperatorApply(RDy rdy, PetscReal dt, Vec U_local, V
...
+    CeedVector   s_ceed       = rdy->ceed.sources;
 
     // 2. Sets the pointer of a CeedVector to a PETSc Vec: u_local_ceed --> U_local
     PetscCall(VecGetArrayAndMemType(U_local, &u, &mem_type));
     PetscCallCEED(CeedVectorSetArray(u_local_ceed, MemTypeP2C(mem_type), CEED_USE_POINTER, u));
 
     // 3. Make a duplicate copy of the F as the values will be used as input for the CeedOperator
     //    corresponding to the source-sink term
-    PetscCall(VecCopy(F, rdy->F_dup));
+    PetscCall(VecCopy(F, rdy->ceed.host_fluxes));
```
bishtgautam, line 286: @jeff-cohere, `rdy->ceed.host_fluxes` should not have included the prefix `host_` because when running on GPUs, this PETSc Vec will be created on the GPU itself as one would use `-dm_vec_type <hip|cuda>`. You can fix this in the next operator refactor PR.

PR #227, opened 2024-10-18
Operator refactor  part 2: moving stuff out of the SWE-specific operators

jeff-cohere:
This PR implements some new data structures for accessing operator data in both CEED and PETSc. There are now specific types for accessing operator data for

* boundary values (fluxes)
* source terms
* material properties
* flux divergences communicated between the flux and source operators

These types pertain to operators for all kinds of conservative flow/transport equations that we might solve, so they can manipulate data for non-SWE operators. Accordingly, they replace their SWE-specific equivalents, allowing operator implementations to focus only on their specific physics. I intend these types also to handle things for both PETSc and CEED runs, whch localizes a lot of logic that differentiates between these two sets of solvers.

The data access structures work very similarly to how data is manipulated in PETSc/CEED vectors. You get access to operator data, set whatever values you want, and then release ("restore") the data in the operators. We allow only one entity at a time to access the data, which is fine, since "host calculations" within an MPI process are done completely serially.

In the next stage of the refactor, I'll attempt to replace SWE PETSc logic with these types, and patch in the flux divergences, which I haven't done yet.


jeff-cohere:
Aha, code coverage! Yep, I added some code that hasn't been used yet. I don't think this is a real problem, since coverage will bounce back up after the next PR.

driver/main.F90
```diff
@@ -152,9 +152,6 @@ program rdycore_f90
       PetscCallA(RDyGetLocalCellZCentroids(rdy_, n, values, ierr))
       PetscCallA(RDyGetLocalCellNaturalIDs(rdy_, n, nat_id, ierr))
 
-      PetscCallA(RDyGetLocalCellManningsNs(rdy_, n, values, ierr))
-      PetscCallA(RDySetManningsNForLocalCells(rdy_, n, values, ierr))
```
bishtgautam, line 156: Will you add the capabilities to get/set Mannings N from a Fortran code in a future PR?

src/operator_data.c
```diff
@@ -0,0 +1,391 @@
...
+    PetscReal *u = boundary_data->storage.petsc.data;
+    for (PetscInt e = 0; e < boundary_data->boundary.num_edges; ++e) {
+      u[e * boundary_data->num_components + component] = boundary_values[e];
+    }
+  }
+
+  PetscFunctionReturn(PETSC_SUCCESS);
+}
+
+PetscErrorCode RestoreOperatorBoundaryData(RDy rdy, RDyBoundary boundary, OperatorBoundaryData *boundary_data) {
```
bishtgautam, line 133: A few functions are missing the comment describing the function, but it can be added in a future PR.

bishtgautam (commented):
This looks good and I have two minor comments.

driver/main.F90
jeff-cohere: This didn't actually do anything, as far as I can tell--it just copied the mannings values out and back in again (right?).

src/operator_data.c
jeff-cohere: Yeah, that sounds good.

driver/main.F90
bishtgautam: Yes, they only extracted or set values for ManningN. These functions could be needed for coupling with E3SM.

bishtgautam: These functions can be added back in a future PR.

bishtgautam (approved):

PR #228, opened 2024-10-21
Adds few destroy functions

bishtgautam:
To support AMR, few destroy functions are added that can be called when the mesh is refined.

### DIFF
--- a/include/private/rdydmimpl.h
+++ b/include/private/rdydmimpl.h
@@ -19,5 +19,6 @@ PETSC_INTERN PetscErrorCode CloneAndCreateCellCenteredDM(DM dm, const SectionFie
 PETSC_INTERN PetscErrorCode CreateDM(RDy rdy);
 PETSC_INTERN PetscErrorCode CreateAuxiliaryDM(RDy rdy);
 PETSC_INTERN PetscErrorCode CreateVectors(RDy rdy);
+PETSC_INTERN PetscErrorCode CreateSectionForSWE(RDy, PetscSection *);
 
 #endif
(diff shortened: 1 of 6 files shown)


include/private/rdydmimpl.h
```diff
@@ -19,5 +19,6 @@ PETSC_INTERN PetscErrorCode CloneAndCreateCellCenteredDM(DM dm, const SectionFie
 PETSC_INTERN PetscErrorCode CreateDM(RDy rdy);
 PETSC_INTERN PetscErrorCode CreateAuxiliaryDM(RDy rdy);
 PETSC_INTERN PetscErrorCode CreateVectors(RDy rdy);
+PETSC_INTERN PetscErrorCode CreateSectionForSWE(RDy, PetscSection *);
```
jeff-cohere, line 22: In the interest of confining SWE-specific code to `rdysweimpl.h` and `src/swe`, maybe we could call this `CreateSection` and have it dispatch to `CreateSWESection` in `rdysweimpl.h` (defined in `src/swe/physics_swe.c`, perhaps).

src/rdycore.c
```diff
@@ -117,6 +117,63 @@ PetscErrorCode RDyCreateF90(MPI_Fint *f90_comm, const char *config_file, RDy *rd
...
+    if ((*rdy)->regions[i].cell_ids) {
+      PetscFree((*rdy)->regions[i].cell_ids);
+    }
+  }
+  if ((*rdy)->regions) PetscFree((*rdy)->regions);
+
+  PetscFunctionReturn(PETSC_SUCCESS);
+}
+
+/// @brief Destroy boundariy data structures
```
jeff-cohere, line 150: boundariy -> boundary

include/private/rdydmimpl.h
bishtgautam: I agree with the change you have suggested, but how about you make that change in the next operator refactor PR?

jeff-cohere (approved):
Looks good to me. I'll tuck the SWE stuff out of sight in the next operator refactor PR.

PR #243, opened 2024-11-15
Separated courant number diagnostic core logic from SWE operator.

jeff-cohere:
This PR pulls some basic logic for our courant number diagnostics (including a MPI datatype and operator for reducing maximum wave speeds) out of the SWE code. The actual determination of the maxwave speed remains part of the SWE code, but as we generalize our approach to "operators", this will also be pulled out.

### DIFF
--- a/src/courant.c
+++ b/src/courant.c
@@ -1,0 +1,63 @@
+#include <private/rdycoreimpl.h>
+
+// MPI datatype corresponding to CourantNumberDiagnostics. Created during
+// CreateSWEOperator.
+MPI_Datatype MPI_COURANT_NUMBER_DIAGNOSTICS = {0};
+
+// MPI operator used to determine the prevailing diagnostics for the maximum
+// courant number on all processes. Created during CreateSWEOperator.
+MPI_Op MPI_MAX_COURANT_NUMBER = {0};
+
+// function implementing the above MPI operator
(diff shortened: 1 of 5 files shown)


bishtgautam (approved):

src/courant.c
```diff
@@ -0,0 +1,63 @@
...
+  static PetscBool initialized = PETSC_FALSE;
+
+  if (!initialized) {
+    // create an MPI data type for the CourantNumberDiagnostics struct
+    const int      num_blocks             = 4;
+    const int      block_lengths[4]       = {1, 1, 1, 1};
+    const MPI_Aint block_displacements[4] = {
+        offsetof(CourantNumberDiagnostics, max_courant_num),
+        offsetof(CourantNumberDiagnostics, global_edge_id),
+        offsetof(CourantNumberDiagnostics, global_cell_id),
```
jedbrown, line 46: I don't think we'll produce these two fields from libCEED, or it would be tricky to do so in an efficient way. Especially in the context of second order methods, I don't understand how both the edge and cell can be uniquely determined.

jeff-cohere: Thanks, Jed. I'm just moving around what we had in there already. I think we'll sort through this stuff when we reorganize our operator representation from our SWE-specific implementation (very soon!).

PR #245, opened 2024-11-16
Changes OperatorSourceData interface to be region-aware.

jeff-cohere:
Last one for the week!

This PR changes the OperatorSourceData interface to associate source data with regions, in the same way as boundary flux data relates to boundaries. The difference between these two cases is that while boundary data is stored per-boundary, all source data lives in one vector of owned-cell data. So we do some index mapping to translate between representations.

**Update**: based on review feedback, I've made a few more changes:

* the `RDyRegion` type now keeps track of IDs of cells for local and global vectors
* the `is_local` field in the `RDyCells` type has been renamed to `is_owned`
* the `is_local` fields in the `RDyVertices` and `RDyEdges` types have been removed

### DIFF
--- a/src/rdydata.c
+++ b/src/rdydata.c
@@ -198,28 +198,17 @@ PetscErrorCode RDySetWaterSourceForRegion(RDy rdy, const PetscInt region_idx, Pe
 
   if (region.num_cells) {
     PetscReal *source_values;
-    PetscCall(PetscCalloc1(rdy->mesh.num_owned_cells, &source_values));
-
-    // get the values for source term associated with h component
-    OperatorSourceData source_data;
-    PetscInt           component = 0;
-    PetscCall(GetOperatorSourceData(rdy, &source_data));
-    PetscCall(GetOperatorSourceValues(&source_data, component, source_values));
-    PetscCall(RestoreOperatorSourceData(rdy, &source_data));
-
-    RDyMesh  *mesh  = &rdy->mesh;
-    RDyCells *cells = &mesh->cells;
-
-    // update the source term for the cells in the region that are local
+    PetscCall(PetscCalloc1(region.num_cells, &source_values));
     for (PetscInt c = 0; c < region.num_cells; c++) {
-      PetscInt cell_id = region.cell_ids[c];
-      if (cells->is_local[cell_id]) {
-        PetscInt owned_cell_id       = cells->local_to_owned[cell_id];
-        source_values[owned_cell_id] = value;
-      }
+      source_values[c] = value;
     }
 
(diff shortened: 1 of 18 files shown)


jeff-cohere:
Sorry for the Friday afternoon PR, Gautam! It can certainly wait till next week--I just wanted to get it off my plate.

src/rdydata.c
```diff
@@ -198,71 +198,76 @@ PetscErrorCode RDySetWaterSourceForRegion(RDy rdy, const PetscInt region_idx, Pe
 
   if (region.num_cells) {
     PetscReal *source_values;
-    PetscCall(PetscCalloc1(rdy->mesh.num_owned_cells, &source_values));
+    PetscCall(PetscCalloc1(region.num_cells, &source_values));
```
bishtgautam, line 201: I think we should allocate memory for `mesh.num_owned_cells`. This is because the source-sink term won't be applied to the ghost cells. Also, we need to check if `region.num_cells` corresponds to the global cells in the region or local+ghost cell.

Btw, we can introduce a new member variable `region.num_owned_cells` and use it to allocate memory.

bishtgautam: I just realized that `mesh.num_owned_cells` isn't the correct choice. We should introduce a new member variable, `region.num_owned_cells`, and use it to allocate memory using it.

bishtgautam (commented):
I think the modification to `<Set/Get>SourceComponentForLocalCells` aren't consistent with what these functions are meant to be. These function are meant to set/get values for all local cells irrespective of the region.

We can add new functions that set/get values for a specific region such as `<Set/Get>SourceComponentForLocalCellsInARegion(source_data, component, region_id, *source_value)`.

Update: Another possibility is to eliminate `<Set/Get>SourceComponentForLocalCells` and only have `<Set/Get>SourceComponentForLocalCellsInARegion`.

jeff-cohere:
> **Update**: Another possibility is to eliminate `<Set/Get>SourceComponentForLocalCells` and only have `<Set/Get>SourceComponentForLocalCellsInARegion`.

I like this approach, given that we associate material properties and conditions with regions in our input spec. I believe we require the entire domain to be covered by the set of regions. I know that Gmsh defines a region that covers the entire domain if we don't specify regions ourself. I don't know if this is true for Exodus, though.

Regarding "owned cells" in regions: I suppose we need to figure out whether cells in regions are owned by definition, or whether PETSc Label includes cells within regions that aren't owned. If all cells in regions are owned, I think it would be good to rename `region.num_cells` to `region.num_owned_cells` to make this clear.

bishtgautam (approved):
Thanks for these changes. It took us a little more work than we expected to get this PR ready.

PR #254, opened 2025-01-13
Added an option to write the grid to a separate file (XDMF only).

### LINKED ISSUE #248: Add an option to write mesh related information in a separate .h5 file
As the mesh becomes larger, it is desirable to output the mesh-related quantities in
a single `.h5` file and not in every `.h5`.

jeff-cohere:
This option lets us save some space when the domain is very large. It's demonstrated in the Simons.OceanDirechletBC.yaml test input.

Closes #248

### DIFF
--- a/docs/common/input.md
+++ b/docs/common/input.md
@@ -406,6 +406,7 @@ output:
   batch_size: 1
   time_series:
     boundary_fluxes: 10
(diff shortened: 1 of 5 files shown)


bishtgautam (approved):
This works perfectly. Thanks.

PR #273, opened 2025-01-29
Adds velocity regularization

bishtgautam:
- Introduces a new parameter, `h_anuga_reg_parameter`, in the .yaml file to 
 support the velocity regularization approach used by the ANUGA hydrodynamic model.

- The velocity in x direction is computed as `u = (hu) * h / (h^2 + h_anuga_regular^2)`. 
  The same approach is used for velocity in y direction.

- The default value of `h_anuga_reg_parameter` is `0.0`.

### DIFF
--- a/docs/common/input.md
+++ b/docs/common/input.md
@@ -440,7 +440,7 @@ physics:
   flow:
     mode: swe
     tiny_h: 1e-7
-    h_anuga_reg_parameter: 1.e-3
+    h_anuga_regular: 1.e-3
     source:
       method: implicit_xq2018
       xq2018_threshold: 1e-10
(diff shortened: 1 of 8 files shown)


bishtgautam:
@donghuix, Can you please review the PR to ensure the ANUGA methodology is correctly implemented?
@jedbrown, Can you please review the PR to check if the vectorization in the CEED code is not broken?
@jeff-cohere, Can you please review the PR for all the other changes in the PR?

docs/common/input.md
```diff
@@ -440,6 +440,7 @@ physics:
   flow:
     mode: swe
     tiny_h: 1e-7
+    h_anuga_reg_parameter: 1.e-3
```
jeff-cohere, line 443: What do you think about calling this `h_anuga_regular`, as you've done in `rdyconfigimpl.h`?

docs/common/input.md
```diff
@@ -460,6 +461,10 @@ parameters in this subsection are:
    not yet supported). This parameter is required and has no default value.
 * `tiny_h`, which is the water height below which a given point is assumed to
   be dry. Default value: `1e-7`
+* `h_anuga_reg_parameter`: is the water height used in velocity regularization using the
```
jeff-cohere, line 464: It's actually the square of a height, right?

jeff-cohere (approved):
This looks good to me. I left a couple very minor comments.

docs/common/input.md
jedbrown: Yeah, I'd rather the definition be $u = (hu)h / (h^2 + h_{\text{anuga}}^2)$ so that the regularization parameter has units of thickness. (This is a deviation from the paper, but I think they chose poor notation.)

donghuix (approved):
The implementation of velocity regularization looks correct to me.

jedbrown (approved):
The qfunction code looks good. When I profiled last, I think the CPU vectorization was not good, which I suspected was related to the Roe solver. We're going to compare with an HLL solver, which I've found to vectorize well in other settings.

docs/common/input.md
bishtgautam: Oh, good point. 

How about I change the code from `denom = Square(data->h[n]) + h_anuga;` to
`denom = Square(data->h[n]) + Square(h_anuga);`?

docs/common/input.md
bishtgautam: Done

PR #276, opened 2025-02-04
Adding support for trapping floating point exceptions.

jeff-cohere:
With these changes, RDycore will trap the most common floating point exceptions in Debug builds using the GNU C compiler. This can help us understand how and where NaNs get generated. I'm using this to unravel the SWE and sediment operators so they can be combined in a way that reduces duplicate code.

### DIFF
--- a/include/private/rdycoreimpl.h
+++ b/include/private/rdycoreimpl.h
@@ -154,6 +154,10 @@ struct _p_RDy {
   PetscViewerAndFormat *output_vf;
 };
 
+// floating point exceptions (GNU-specific)
+PETSC_INTERN PetscErrorCode EnableFloatingPointExceptions(void);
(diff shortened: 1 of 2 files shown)


jeff-cohere:
GCC offers some extensions to the C standard that allow for easy floating point exception tracking. I'm not aware of a way to do this with Apple's compiler, but if I figure that out and we want to trap these exceptions, I'll add it in. I've tested this feature in my composable operator work and it does successfully trap FPEs.

bishtgautam (approved):

jeff-cohere:
Hmmm. Trapping floating point exceptions is always a tricky game in HPC. Lots of tests are failing for benign reasons, so this might not be worth doing after all.

PR #277, opened 2025-02-07
Composable Operators, Part 1: Disentanglement

jeff-cohere:
This PR rewires the SWE and SD (sediment dynamics) flux/source operators so that they can be stacked atop one another, with each one implementing only one set of physics.

This might seem a little abstract, so let's get specific: before this PR, our "sediment" flux and source operators actually implemented both SWE and sediment physics. Now these sediment operators only implement the sediment part, and both SWE and sediment operators are combined to solve the coupled system.

Untangling these operators required several changes which fortunately seem to clean things up:

* Source, flux divergence, and boundary operator data must be shared between stackable operators
* All material properties are now stored within a single PETSc or CEED vector
* SWE and SD operators have had their math cleaned up a bit.

Now that we have working composable operators, it's time to reorganize things and remove some unnecessary code. But because these operators can be crafted for introducing new physics, they are likely an area of growth in RDycore that can be undertaken by multiple contributors. It's a good time for us to talk about how to organize their development.

### DIFF
--- a/docs/theory/sediment.md
+++ b/docs/theory/sediment.md
@@ -98,7 +98,7 @@ clay and 1.13 for loam.
 #### Overland flow-driven entrainment and re-entrainment rates
 
 \begin{eqnarray}
-r_p &=& (1-H)f_p}\frac{F(\Omega-\Omega_{cr})}{J} \\
+r_p &=& (1-H)f_p\frac{F(\Omega-\Omega_{cr})}{J} \\
 r_{rp} &=& H\frac{M_p}{M_{t}}\frac{F(\Omega - \Omega_{cr})}{(\rho_{s}-\rho_{w})gh/\rho_{s}} \tag{6}\label{6}
 \end{eqnarray}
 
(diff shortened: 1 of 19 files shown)


jeff-cohere:
Some useful links from @jedbrown regarding "composable Q functions" instead of "composable operators":

* [Example of how to use enums within Q functions to dispatch work to helper functions (e.g. for switching between different Riemann solvers](https://gitlab.com/phypid/honee/-/blob/main/qfunctions/bc_freestream.h?ref_type=heads#L119-141)
* [Example of using helper functions to reuse code within operators (e.g. to simplify the SWE + SD operator, as opposed to combining SWE and SD operators](https://gitlab.com/micromorph/ratel/-/blob/main/include/ratel/qfunctions/models/elasticity-neo-hookean-damage-common.h?ref_type=heads#L96)

bishtgautam (approved):
The `operator.c` is getting larger and we may split create additional `operator_ceed` and `operator_petsc` files. But, let's wait till the Part 2 of this PR.

jeff-cohere:
I think this PR contains a lot of good changes, but the approach diverges a bit from what Jed has suggested. I'm going to close it and move the stuff we want to keep to a new branch. In this new branch, I'll reorganize the Q functions and helpers into a reusable set for a single set of flow/source operators, similar to how SD is implemented in `main`.

Issue #278, opened 2025-02-12
Reorganize Q functions and helpers to provide building blocks for flow + sediment + salinity operators


jeff-cohere:
In #277 we explored some ideas on how to make "composable operators" to mix and match physics, now that we're dealing with coupled flow (SWE) and sediment dynamics (SD) models. In our last meeting, I asked @jedbrown about how such compositions are done in other CEED codes, and he gave us some useful links that suggest we should compose Q functions and supporting helper functions instead of the operators themselves:

* [Example of how to use enums within Q functions to dispatch work to helper functions (e.g. for switching between different Riemann solvers)](https://gitlab.com/phypid/honee/-/blob/main/qfunctions/bc_freestream.h?ref_type=heads#L119-141)
* [Example of using helper functions to reuse code within operators (e.g. to simplify the SWE + SD operator, as opposed to combining SWE and SD operators)](https://gitlab.com/micromorph/ratel/-/blob/main/include/ratel/qfunctions/models/elasticity-neo-hookean-damage-common.h?ref_type=heads#L96) 

## TODO

* [x] Close #277 
* [x] Create a new branch and migrate to it all useful changes from #277 **<-- see [`jeff-cohere/q-function-reorg`](https://github.com/RDycore/RDycore/tree/jeff-cohere/q-function-reorg)**
* [x] Reorganize Q functions and helpers to enable composition **<-- see #282**
* [x] Consolidate flow and source operator code to select the correct Q functions for a specific configuration **<-- see #280**
* [x] Create new PR for discussion, approval

PR #279, opened 2025-02-13
Minor operator cleanup before Q-function reorganization

jeff-cohere:
This PR brings in some useful improvements from #277. There's no reason not to merge these before starting work on #278.


jeff-cohere:
@bishtgautam , you've seen all of this code before, so it should be an easy review.

bishtgautam (approved):
Looks good.

PR #280, opened 2025-02-13
CEED Q function reorganization, part 1: consolidating operator constructors

jeff-cohere:
This PR separates the PETSc and CEED operator constructor logic into separate files. It also gathers CEED operator constructors into a single location (`src/operator_ceed.c`), separating only Q function and helper files. The idea here is that `src/operator_ceed.c` contains all logic for constructing a CEED operator that is appropriate for a physical/numeric configuration of interest. I've organized the creation of the Q function and its context according to which physics is supposed to be enabled.

When you want to add new physics to a CEED operator, you write new Q functions (and perhaps helpers) and then modify logic in `src/operator_ceed.c` to use these Q functions according to the appropriate configuration fields.

NOTE: I haven't refactored the sediment dynamics Q functions to use helpers for computing flow terms yet, so there's still code duplication there. In the next PR I will refactor those Q functions. In subsequent PRs, I'll reorganize the PETSc operators to look similar to the CEED operators.


bishtgautam (approved):
Thanks for moving some code out of the giant operator.c.

PR #281, opened 2025-02-14
Fixes for running the Amazon lake problem in parallel

bishtgautam:
This PR includes two fixes:
1. A source term that might be applied to a region that is absent on a given MPI rank
   due to domain decomposition. In this case, we skip the application of the source
   value for the region.
2. After domain decomposition, a rank may have no boundary cells. So, we need
    only apply the code to identify boundary cells is a boundary edge has been identified.


jeff-cohere (approved):
Looks great! I agree that it's better to use the (1-based) region IDs from the mesh files than the (0-based) indices of regions as they're stored internally.

PR #282, opened 2025-02-18
CEED Q function reorganization, part 2: refactoring SWE and sediment Q functions

jeff-cohere:
This PR breaks up our existing Q functions to allow us to reuse certain helper functions, and to isolate Riemann-solver-specific logic to certain files.

After this, I'll do a quick cleanup of the PETSc SWE/sediment operators to give it a similar structure.

Closes #278

PR #283, opened 2025-02-19
Reorganized PETSc SWE and sediment operators to resemble reorganized CEED operators.

jeff-cohere:
With this PR, we now have a completely reorganized set of SWE and sediment operators, with separate files for different Riemann solvers. My next task is to document our redesigned operators to make it easier for the team to work with them.

This PR is in line behind #282 .


bishtgautam (approved):
This looks good.

PR #286, opened 2025-03-04
Adds a shell script for configuring a ELM-RDycore on Perlmutter for ROS event

bishtgautam:
This adds the script for configuring the 1996 Pacific Northwest ROS as
ELM-RDycore configuration. Currently, this configuration is supported
for Perlmutter's CPU and GPU nodes.


bishtgautam:
And here is a sample snapshot of RDycore simulated water height at the end of the first day of the simulation.

water_height

daleihao (approved):

PR #287, opened 2025-03-04
Updates license

bishtgautam:
This updates the license to list the institutions explicitly.


jeff-cohere (approved):

jedbrown (approved):
It's "fine", but then our contributing guidelines should then state that every contributor should add their name (or copyright-holding institution) to the list. That also means that if a student contributes as part of a class project, they add their name. This is generally considered unwieldy, and is why pretty much zero open source projects do it.

jonbob (approved):

markadams (approved):

knepley (approved):

bishtgautam:
For the record, @MukeshKumar-Hydrology approved this change via email.

Issue #299, opened 2025-04-04
With PETSc v3.23.0, RDycore build hangs while building Fortran modules


bishtgautam:
Based on the conversation on Slack

After switching to the latest PETSc `95934b0d3930a39ae37491dd05d108b6eb525436`, the compiling of RDycore is stuck in scanning for Fortran dependency. I was able to disable compiling Fortran code in RDycore via:

```
diff --git a/driver/CMakeLists.txt b/driver/CMakeLists.txt
index 959fc0b..215c4f2 100644
--- a/driver/CMakeLists.txt
+++ b/driver/CMakeLists.txt
@@ -25,25 +25,25 @@ install(
 )
 
 # Fortran driver
-add_executable(rdycore_f90_exe
-  main.F90
-)
-set_target_properties(rdycore_f90_exe PROPERTIES OUTPUT_NAME rdycore_f90)
-target_include_directories(rdycore_f90_exe PRIVATE ${PROJECT_BINARY_DIR}/src/f90-mod)
-target_link_libraries(rdycore_f90_exe PRIVATE rdycore_f90)
+#add_executable(rdycore_f90_exe
+#  main.F90
+#)
+#set_target_properties(rdycore_f90_exe PROPERTIES OUTPUT_NAME rdycore_f90)
+#target_include_directories(rdycore_f90_exe PRIVATE ${PROJECT_BINARY_DIR}/src/f90-mod)
+#target_link_libraries(rdycore_f90_exe PRIVATE rdycore_f90)
 
 # MMS Fortran driver
-add_executable(rdycore_mms_f90
-  mms.F90
-)
-set_target_properties(rdycore_mms_f90 PROPERTIES OUTPUT_NAME rdycore_mms_f90)
-target_include_directories(rdycore_mms_f90 PRIVATE ${PROJECT_BINARY_DIR}/src/f90-mod)
-target_link_libraries(rdycore_mms_f90 PRIVATE rdycore_f90)
+#add_executable(rdycore_mms_f90
+#  mms.F90
+#)
+#set_target_properties(rdycore_mms_f90 PROPERTIES OUTPUT_NAME rdycore_mms_f90)
+#target_include_directories(rdycore_mms_f90 PRIVATE ${PROJECT_BINARY_DIR}/src/f90-mod)
+#target_link_libraries(rdycore_mms_f90 PRIVATE rdycore_f90)
 
-install(
-  TARGETS rdycore_mms_f90 rdycore_f90_exe
[...]

PR #302, opened 2025-04-08
Adding some workarounds for Fortran-related difficulties

jeff-cohere:
This PR adds an `ENABLE_FORTRAN` switch that is `ON` by default but that can be turned `OFF` to disable Fortran 90 bindings. CMake is having issues (#299) generating dependencies for PETSc's new Fortran bindings, and this PR unblocks team members who don't rely on Fortran.

@jedbrown , you can cherry-pick commit `c05f909` to get this switch, and add `-DENABLE_FORTRAN=OFF` to your configuration command. I'm going to see if I can get [Ninja](https://ninja-build.org/) working as an alternative to Make, since it seems to follow a different code path for generating dependencies and doesn't exhibit the hang we're observing.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,6 +2,7 @@ cmake_minimum_required (VERSION 3.14.0)
 
 # Configurable options
 option(ENABLE_DRIVER   "Enable stand-alone RDycore driver program(s)" ON)
+option(ENABLE_FORTRAN  "Enable Fortran library bindings" ON)
 option(ENABLE_TESTS    "Enable unit and driver tests" ON)
 option(ENABLE_COVERAGE "Enable code coverage instrumentation" OFF)
 option(ENABLE_SANITIZERS "Enable available sanitizers (ASan, MSan, UBSan)" OFF)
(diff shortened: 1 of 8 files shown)


jedbrown:
Oh, interesting. Yeah, I can build with no changes if I use `cmake .. -G Ninja`. (I use Ninja for big projects like building LLVM so that doesn't bother me as a workaround.)

jeff-cohere:
Using Ninja instead of Make works for me too. @bishtgautam , you can get Ninja using Homebrew, and it's essentially a drop-in replacement for Make (e.g. `ninja` builds the code, `ninja test` runs tests, etc). This is a nice workaround for now.

Using PETSc's `main` branch, I had to replace calls to `Vec{Get,Restore}ArrayF90` with `Vec{Get,Restore}Array` to link properly. Once we decide on a new version of PETSc to use for our Docker images, I can add those changes to this branch and we can switch over in this PR.

jeff-cohere:
Okay, it works!

@bishtgautam , I'm adding one more commit that explains how to use Ninja instead of Make, but it's only documentation, so it shouldn't affect the tests.

bishtgautam (approved):
Using the following command, we can now install RDycore on Perlmutter (with 6 versions of PETSc) and Frontier (with 3 versions of PETSc)

```
cmake .. -DCMAKE_INSTALL_PREFIX=$PWD -DCMAKE_BUILD_TYPE=Debug -G Ninja
ninja -j4 install
```

Thanks, Jeff.

jeff-cohere:
Thanks for verifying on the Big Iron, @bishtgautam !

PR #306, opened 2025-05-01
Updates the Fortran driver

### LINKED ISSUE #305: Update Fortran driver to support unstructured BC dataset
In preparation for the ELM-RDycore flooding study during Irene, update the Fortran driver to support the BC dataset in the unstructured grid format, which the C driver currently supports.

bishtgautam:
Updates the Fortran driver to support the following:
- Rainfall dataset in the raster and unstructured grid format.
- Boundary condition dataset in spatially homogeneous and unstructured grid format.

Closes #305


jeff-cohere:
Yes, I think what you suggest would work.

We could also just `#include <config.h>` in `include/finclude/rdycore.h` with some minor adjustments, if we're willing to remove all the C comments in `config.h.in` and document these parameters elsewhere. I've figured out the changes that would be needed if you would rather just use the one generated `config.h` file for both C and Fortran.

driver/main.F90
```diff
@@ -74,6 +194,1263 @@ subroutine getcurrentdata(data_ptr, ndata, cur_time, temporally_interpolate, cur
...
+          ndays = 29
+        else
+          ndays = 28
+        end if
+    end select
+
+  end subroutine day_of_month
+
+  ! Adds one hour to current date and updates the date if necessary
+  subroutine IncrementDateByOneHour(current_date)
```
jeff-cohere, line 586: Are you convinced that this is correct? We could wrap our `mktime` C solution if this isn't reliable.

jeff-cohere (approved):
This looks good, at least till we incorporate datasets into the library. I left one minor question.

driver/main.F90
bishtgautam: For the record: Since we will fold these new data structures into the RDycore lib itself in a future PR, we decided to leave the code unchanged.

PR #309, opened 2025-05-05
Adding the ability to select which fields (components) appear in output

### LINKED ISSUE #294: Add an output subsection indicating fields to be dumped
Currently, an output file includes the mesh geometry and our solution field components. We would like to be able to dump other field data (sources, for example). We can add a subsection to the `output` section (`fields`, perhaps?) that lists the desired output fields. If this subsection is omitted, RDycore should output only the solution fields as is done now.

## Notes

Currently, we store fields in 2 DMs:

* a "primary" DM (`rdy->dm`), which contains a `Solution` field with a set of components relevant for the desired physics:
    * `Height`, `MomentumX`, `MomentumY` for the shallow water equations
    * `Concentration_<I>` for sediment class `<I>` in the case of sediment dynamics

* an "auxiliary" DM (`rdy->aux_dm`), which so far doesn't contain much:
    * `Parameter`, a single-component field, in the case of the RDycore driver
    * `Error`, the error vector for the `Solution` field in the primary DM, in the case of the MMS driver

I assume that this selective output feature would be most useful for the main driver.

## Questions

1. **Should we put non-solution fields in the auxiliary DM?**
2. **What non-solution fields do we want?** I assume source terms, such as `WaterSource`, `MomentumXSource`, `MomentumYSource`, `Concentration_<I>Source` (perhaps). What else?

jeff-cohere:
This PR adds a `fields` subsection to the `output` section of the input specfication. This new subsection contains a list of desired output field components for the solution and diagnostic vectors. By default, all components of the solution vector appear in output. More details appear in the documentation.

I made a few adjustments to logic to try to make this less confusing, and I've tested it a little bit, but I welcome more of that! :-)

Closes #294

### DIFF
--- a/docs/common/input.md
+++ b/docs/common/input.md
@@ -424,7 +424,8 @@ time series data (**but excluding checkpoint data**). Relevant parameters are
   array syntax (typically with hyphenated items on separate lines). This item is
   optional, and if omitted the selected fields are the components of the
   solution (e.g. `Height`, `MomentumX`, `MomentumY` for the shallow water
-  equations). Available options are
+  equations). **NOTE: This feature works only for XDMF output!** Available
+  options are
     * `Height`: water height $h$
     * `MomentumX`: $x$ momentum $hu$
     * `MomentumY`: $y$ velocity $hv$
--- a/docs/common/input.md
+++ b/docs/common/input.md
@@ -407,7 +407,7 @@ output:
     - MomentumY
     - WaterSource
   format: xdmf
-  step_interval: 100
+  output_interval: 100
   batch_size: 1
   time_series:
     boundary_fluxes: 10
@@ -439,7 +439,7 @@ time series data (**but excluding checkpoint data**). Relevant parameters are
     * `binary`: output is written using PETSc's binary data format
     * `xdmf`: output is written to the [XDMF](https://xdmf.org/index.php/XDMF_Model_and_Format) format
     * `cgns`: output is written to the [CFD General Notation System (CGNS)](https://cgns.github.io/) format
-* `step_interval`: the number of time steps between output dumps. Default value: 0 (no output)
+* `output_interval`: the number of time steps between output dumps. Default value: 0 (no output)
 * `time_interval`: the temporal frequency between output dumps. The is an integer with a minimum value of 1 second. Default value: 0 (no output)
 * `time_unit`: units of temporal frequency output.
 * `batch_size`: the number of time steps for which output data is stored in a
--- a/docs/common/input.md
+++ b/docs/common/input.md
@@ -401,6 +401,8 @@ RDycore. The parameters that define these discretizations are
 ```yaml
 output:
   directory: output
+  fields:
+    - 
   format: xdmf
   step_interval: 100
   batch_size: 1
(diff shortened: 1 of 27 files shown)


bishtgautam:
@jeff-cohere, I tested a case in which rainfall is non-zero, but the output in the .h5 is for `WaterSource` is zero.

- I modified `Houston1km.DirichletBC.yaml` so it only runs for a single time step, and model writes the output.
```
time:
  final_time       : 30
  time_step        : 30.0
  coupling_interval: 30.0
  unit             : seconds

output:
  format: xdmf
#  step_interval: 10000
  time_interval: 30
  time_unit: seconds
  fields:
    - Height
    - MomentumX
    - MomentumY
    - WaterSource
  batch_size: 20
  time_series:
    boundary_fluxes: 1
```

- Here is the code modification to show that rain is non-zero
```
diff --git a/driver/main.c b/driver/main.c
index 0b9d302..e3d292d 100644
--- a/driver/main.c
+++ b/driver/main.c
@@ -1050,6 +1050,7 @@ PetscErrorCode SetHomogeneousData(HomogeneousDataset *homogeneous_rain, PetscRea
   PetscInt    *cur_rain_idx           = &homogeneous_rain->cur_idx;
   PetscInt    *prev_rain_idx          = &homogeneous_rain->prev_idx;
   PetscCall(GetCurrentData(rain_ptr, nrain, cur_time, temporally_interpolate, cur_rain_idx, &cur_rain));
+  printf("cur_rain = %e\n",cur_rain);

   if (temporally_interpolate || *cur_rain_idx != *prev_rain_idx) {  // is it time to update the source term?
     *prev_rain_idx = *cur_rain_idx;
```

- I ran the RDycore driver as:
```
cd build/driver/tests/swe_roe
$PETSC_DIR/build-$PETSC_ARCH/bin/mpiexec
 ../../rdycore   \
 Houston1km.DirichletBC.yaml   \
 -homogeneous_rain_file   Houston1km.rain.int32.bin  \
 -homogeneous_bc_file   Houston1km.bc.int32.bin   \
-temporally_interpolate_bc
```

jeff-cohere:
Thanks for testing, Gautam. I'll take a look.

src/yaml_input.c
```diff
@@ -234,9 +234,14 @@ static const cyaml_schema_field_t restart_fields_schema[] = {
 // output section
 // ---------------
 // output:
+//   directory: <output-directory>
+//   fields: <list-of-output-fields>
 //   format: <binary|xdmf|cgns>
-//   interval: <number-of-steps-between-output-dumps> # default: 0 (no output)
+//   step_interval: <number-of-steps-between-output-dumps> # default: 0 (no output)
```
bishtgautam, line 240: How about `output_interval`?

docs/common/input.md
```diff
@@ -401,6 +401,11 @@ RDycore. The parameters that define these discretizations are
 ```yaml
 output:
   directory: output
+  fields:
```
bishtgautam, line 404: How about updating `Houston1km.DirichletBC.yaml` instead of this one, as we will get non-zero values for `WaterSource`?

bishtgautam (approved):
Looks good.

PR #310, opened 2025-05-07
Replacing our existing clang-format GitHub Action with a new one that shows nonconforming files.

### LINKED ISSUE #260: Add formatting diffs to CI code formatting check
From @knepley in #259 : 

> This fails the clang format check, but it does not give me a diff. I have a different version of clang-format so I cannot run it myself. Can we set it up to give back the diff? PETSc does this so I can manually fix it.

Good idea. But it's going to take a little fiddling, so here's an issue.

jeff-cohere:
Closes #260

### DIFF
--- a/.github/workflows/clang-format-check.yml
+++ b/.github/workflows/clang-format-check.yml
@@ -5,9 +5,9 @@ jobs:
     name: C code formatting check
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v3
-    - name: Check C formatting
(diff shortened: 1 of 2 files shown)


jeff-cohere:
[Here's](https://github.com/RDycore/RDycore/actions/runs/14892018530/job/41826113380?pr=310) an example of what happens when improperly formatted code is commited.

jedbrown (approved):

bishtgautam (approved):

PR #327, opened 2025-07-31
Adds some default output fields.

### LINKED ISSUE #326: A few settings are missing in default .xmf
If no variables are explicitly asked for output, the `.xmf` file does not include Height/MomentumX/MomentumY in the output. However, by default, h/hu/hv is written out in `.h5` file. So, let's add h/hu/hv by default.

### LINKED ISSUE #329: The DataItem description for XC/YC/ZC in .xmf is wrong
The .xmf has
```
      <Attribute Name="XC" AttributeType="Scalar" Center="Cell">
        <DataItem Dimensions="928560" Format="HDF">
          ca_ar_1k.rdycore_offline-0.h5:/fields/XC
        </DataItem>
      </Attribute>

```
while the data in `ca_ar_1k.rdycore_offline-0.h5` is in `Domain/ZC` instead of `fields/ZC`.

jeff-cohere:
Closes #326
Closes #329

### DIFF
--- a/src/rdydm.c
+++ b/src/rdydm.c
@@ -4,6 +4,20 @@
 #include <private/rdydmimpl.h>
 #include <rdycore.h>
 
+static PetscErrorCode RenameDMFields(DM dm, SectionFieldSpec fields) {
+  PetscFunctionBeginUser;
+
+  PetscSection sec;
+  PetscCall(DMGetLocalSection(dm, &sec));
+  for (PetscInt f = 0; f < fields.num_fields; ++f) {
+    PetscCall(PetscSectionSetFieldName(sec, f, fields.field_names[f]));
+    for (PetscInt c = 0; c < fields.num_field_components[f]; ++c) {
+      if (fields.field_component_names[f][c][0]) PetscCall(PetscSectionSetComponentName(sec, f, c, fields.field_component_names[f][c]));
+    }
+  }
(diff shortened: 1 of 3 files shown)


jeff-cohere:
Hmm. Now it seems like only grid data is being dumped to the HDF5 files (at least in the case of the failing tests). I think I have to understand a bit more about what's going on.

I'm out of town Monday and Tuesday next week, but I'll return to this on Wednesday and give it a closer look.

knepley:
@jeff-cohere Crap! That is a bug. Put it on the list.

jeff-cohere:
> @jeff-cohere Crap! That is a bug. Put it on the list.

Interesting (and a relief for my own brain)! Do you need anything other than what's here to construct a reproducer?

bishtgautam (approved):
The fixes work perfectly. Thanks.

src/rdydm.c
```diff
@@ -157,6 +164,9 @@ PetscErrorCode CreateDM(RDy rdy) {
 
   PetscCall(DMViewFromOptions(rdy->dm, NULL, "-dm_view"));
 
+  // rename the fields in the distributed section
```
jeff-cohere, line 167: @knepley , here's where we rename the fields after the distribution. We can remove this when the bug you mentioned is fixed.

PR #334, opened 2025-08-27
CMAKE_MODULE_PATH is now set in an E3SM-compatible way.

jeff-cohere:
This is an easy one! :-)

With this change, E3SM is able to find and use RDycore-supplied CMake functions and macros.


bishtgautam (approved):
Thanks, @jeff-cohere

PR #337, opened 2025-09-04
Adds sample MATLAB code for preprocessing rainfall data

bishtgautam:
Adds MATLAB scripts to convert raster ASCII rainfall data into:
- netcdf files that E3SM's DLND can use, and
- binary files that RDycore can use.


tools/preprocessing/spatially-distributed-rainfall/convert_asc_rainfall_data_to_dlnd_netcdf_files.m
```diff
@@ -0,0 +1,126 @@
...
+
+end
+
+% After checking with Donghui, an offset of cellsize/2 was not added in the 
+% X/Y values
+X = xlc : cellsize : xlc + (ncols-1)*cellsize;
+Y = ylc : cellsize : ylc + (nrows-1)*cellsize;
+[X,Y] = meshgrid(X,Y);
+
+proj = projcrs(32610);
```
fdongyu, line 106: This call and that below requires the Mapping Toolbox, which may not be available on all systems. I cannot run it on Perlmutter. But I think this is fine since the script is used for demonstration purposes.

bishtgautam:
@fdongyu, thanks for the comments. I have updated the PR and re-requested your review.

fdongyu (approved):

PR #342, opened 2025-09-29
Removes a target-related conflict between EAMxx and RDycore in E3SM cases.

jeff-cohere:
Specifically: we now check to see whether the `memcheck` CMake target exists before creating it.

@bishtgautam , I've verified that the case you mentioned now builds successfully when the RDycore submodule contains this change, so after this is merged we can fast-forward the RDycore submodule in the appropriate RDycore/E3SM branch.


bishtgautam (approved):
This is great :)

PR #343, opened 2025-10-03
fix dry cells conditions at boundary

donghuix:
Fix bug for the dry cells at boundary cells. 
Run simulation with dry initial condition and ocean boundary condition, and the water can inundate from boundary condition into the domain.


bishtgautam (commented):
Can you also please fix the same bug in the SD code at https://github.com/RDycore/RDycore/blob/main/src/sediment/sediment_ceed_impl.h#L128?

donghuix:
> Can you also please fix the same bug in the SD code at https://github.com/RDycore/RDycore/blob/main/src/sediment/sediment_ceed_impl.h#L128?

Fixed.

bishtgautam (approved):

PR #351, opened 2025-10-16
Adds initial infrastructure for doing edge reconstruction based on VFR

bishtgautam:
This PR adds the initial infrastructure to support edge-based reconstruction
based on Volume-Free Surface Area relationships for partially wet-dry cells.


bishtgautam:
This isn't needed at the moment, and we will revisit this in the future.

PR #356, opened 2025-11-11
Reorganizing SWE and Sediment operator source to accommodate implement treatment of bed friction source terms

jeff-cohere:
As part of the introduction of our ARK-IMEX time integration capability, I'm breaking up some files and renaming them to indicate more clearly where things live. This is only a source reorganization and doesn't alter the workings of RDycore in any way.

Because I'm working with a newer version of CEED that requires a newer version of PETSc, I've also made minor adjustments to our F90 module that allow things to work with both the version we have and the new version.

### DIFF
--- a/src/f90-mod/rdycore.F90
+++ b/src/f90-mod/rdycore.F90
@@ -5,7 +5,7 @@
 
 module rdycore
 
-  use iso_c_binding, only: c_ptr, c_int, c_int64_t, c_loc
+  use iso_c_binding
   use petscsys
 
   implicit none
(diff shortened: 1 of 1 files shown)


bishtgautam (approved):
Renaming the CEED files and splitting them is a good thing. Thanks.

bishtgautam:
@jeff-cohere what is the new version of PETSc that you are using? I have to reinstall PETSc on perlmutter. So, I can install the version of PETSc that will be required for us in a couple of weeks.

PR #366, opened 2026-01-22
Adds correction term of Bradford-Sanders 2002: Take-2

bishtgautam:
The correction term described in equation 13 of Bradford and Sanders (2002) is
implemented.

Bradford, Scott F., and Brett F. Sanders. "Finite-volume model for shallow-water flooding
of arbitrary topography." Journal of hydraulic engineering 128.3 (2002): 289-298.

PR #369, opened 2026-01-30
Fixes BC flux data for reflective and outflow BC

bishtgautam:
- Need to set values in variables instantaneous and accumulative fluxes for reflective BC.
- For the outflow BC, accumulate zero flux for the case water height is below the threshold.


bishtgautam:
PR #366 replaces this one.

PR #380, opened 2026-03-16
Adds Hydrostatic Reconstruction

bishtgautam:
### Small domain
small_domain_3d

### Initial condition for lake-at-rest
h_ic

### Default model fails to preserve lake-at-rest after 60 secs with dt = 0.1 sec
h_default

### The Hydrostatic Reconstructions preserves lakt-at-rest
h_hr


bishtgautam:
@knepley, This PR contains that sample files that you can use to extend DMPlex to return cell-centered Z values without computing them based on the Z value at the vertices of cells.

- The mesh exo file is `share/meshes/levee.exo`
- The cell-centered Z values in PETSc Binary Vecs are at

```
share/meshes/levee_elevation.int32.bin 
share/meshes/levee_elevation.int64.bin 
```

Let me know if you need additional files.

include/private/rdyoperatorimpl.h
```diff
@@ -191,12 +191,18 @@ PETSC_INTERN PetscErrorCode ApplyOperator(Operator *, PetscReal, Vec, Vec);
 //--------------------------------------------------
 
 PETSC_INTERN PetscErrorCode CreateCeedFluxOperator(RDyConfig *, RDyMesh *, PetscInt, RDyBoundary *, RDyCondition *, CeedVector *, CeedOperator *);
+PETSC_INTERN PetscErrorCode CreateCeedFluxHydroReconOperator(RDyConfig *, RDyMesh *, PetscInt, RDyBoundary *, RDyCondition *, CeedOperator *);
+PETSC_INTERN PetscErrorCode CreateCeedBoundaryFluxSuboperator(const RDyConfig, RDyMesh *, CeedVector *, RDyBoundary *, RDyCondition, CeedOperator *);
 PETSC_INTERN PetscErrorCode CreateCeedSourceOperator(RDyConfig *, RDyMesh *, CeedOperator *);
+PETSC_INTERN PetscErrorCode CreateCeedSourceHydroReconOperator(RDyConfig *, RDyMesh *, CeedOperator *);
```
jeff-cohere: Since you use `HR` to indicate hydrostatic reconstruction elsewhere, what do you think about renaming this to `CreateCeedSourceHROperator`?

src/rdysetup.c
```diff
@@ -1448,6 +1470,12 @@ PetscErrorCode RDySetup(RDy rdy) {
 
   RDyLogDebug(rdy, "Creating FV mesh...");
   PetscCall(RDyMeshCreateFromDM(rdy->dm, 0, &rdy->mesh));
+  if (rdy->config.physics.flow.well_balancing == WELL_BALANCING_HR) {
+    PetscCall(RDyMeshOverride2DProjection(&rdy->mesh));
+  }
+  if (rdy->config.grid.cell_elevation.file[0]) {
+    PetscCall(OverrideCellElevation(rdy));
+  }
```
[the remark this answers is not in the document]
jeff-cohere: Do these overridden cell elevations affect the actual geometry of the mesh, or are they somehow only used in the hydrostatic reconstruction? If they affect the mesh's proper geometry, we need to recalculate the mesh geometry after overriding the elevations as indicated here.

src/swe/swe_fluxes_ceed.h
```diff
@@ -36,9 +36,12 @@ CEED_QFUNCTION_HELPER CeedScalar ComputeDhv(CeedScalar zv_beg, CeedScalar zv_end
 
 // supported Riemann solver types
 #include "swe_roe_flux_ceed.h"
+#ifndef RIEMANN_FLUX_TYPE_DEFINED
+#define RIEMANN_FLUX_TYPE_DEFINED
 typedef enum {
```
jeff-cohere, line 41: If we stick this into `swe_types_ceed.h`, we can avoid these guards and remove this struct from `swe_fluxes_hydro_recon_ceed.h`, right?

src/swe/swe_fluxes_hydro_recon_ceed.h
```diff
@@ -0,0 +1,160 @@
...
+#include "swe_roe_flux_ceed.h"
+#ifndef RIEMANN_FLUX_TYPE_DEFINED
+#define RIEMANN_FLUX_TYPE_DEFINED
+typedef enum {
+  RIEMANN_FLUX_ROE,
+} RiemannFluxType;
+#endif
+
+// geom component indices for interior edges (6 components)
+#ifndef HR_INTERIOR_GEOM_INDEX_DEFINED
```
jeff-cohere, line 39: Consider placing this into `swe_ceed_types.h`.

jeff-cohere (approved):
There's quite a bit of code duplication here, but I think we can probably reorganize it more systematically when we identify the primary configurations we want to support.

I left a couple of suggestions about naming and where to declare some types.

src/rdysetup.c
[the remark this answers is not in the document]
bishtgautam: These overridden cell elevations don't actually affect the actual geometry of the mesh.

src/swe/swe_fluxes_ceed.h
bishtgautam: I have gotten myself confused while trying to move these structs to a different file. So, for now let's leave the code as is and we will revisit it in the future.

PR #376, opened 2026-02-27
Fixed a JIT-related problem introduced by the tracer reorganization PR

jeff-cohere:
Well, that was fun. Maybe I'll learn to control myself better in these large PRs.

The [CEED vector error](https://github.com/RDycore/RDycore/issues/373) reported by @bishtgautam was caused by me moving a generated header file needed by CEED's JIT compiler. The JIT compile failed, initiating a fallback process that exhibits read/write contention in a work vector. I've left a note in `.gitignore` that should dissuade me and others from trying to move generated CEED files from the directory in which they're needed.

Because I can't completely control myself, I also fixed a couple of other little issues I found.

Closes #373

Issue #373, opened 2026-02-22
RDycore build broken on Perlmutter and also getting a CEEDVector error


bishtgautam:
@jeff-cohere, 

1. The latest build is giving the following error on Perlmutter
```
ninja -j4 install
[0/1] Re-running CMake...
-- Cray Programming Environment 2.7.32 C
-- NOTE: LOADEDMODULES changed since initial config!
-- NOTE: this may cause unexpected build errors.
-- Cray Programming Environment 2.7.32 CXX
-- Cray Programming Environment 2.7.32 Fortran
-- PETSc is configured with 32-bit indices
-- Extracted PETSc variable: MPIEXEC = /usr/bin/srun -G4
-- Configuring RDycore (Debug)
-- Generating project files in build directory: /global/cfs/projectdirs/m4267/gbisht/rdycore/build-pm-gpu-hdf5_1_14_3-opt-32bit-gcc-13-2-1-b522cb8c110
-- C compiler is cc (GNU 13.2.1)
-- Fortran compiler is ftn (GNU 13.2.1)
-- Configuring RDycore v0.1.0 (git hash 2ddf02cf)
-- Installation prefix is /global/cfs/projectdirs/m4267/gbisht/rdycore/build-pm-gpu-hdf5_1_14_3-opt-32bit-gcc-13-2-1-b522cb8c110
-- Generating /global/cfs/projectdirs/m4267/gbisht/rdycore/build-pm-gpu-hdf5_1_14_3-opt-32bit-gcc-13-2-1-b522cb8c110/include/cyaml_sediment_fields.h
-- Extracted PETSc variable: BUILDSHAREDLIB = yes
CMake Error at cmake/extract_petsc_variable.cmake:11 (message):
  Could not extract from PETSc.  Please set using -D=...
Call Stack (most recent call first):
  cmake/extract_mpi_include_directories.cmake:6 (extract_petsc_variable)
  CMakeLists.txt:219 (extract_mpi_include_directories)
```
AI agent suggested the following fix that works, but it you should check if that make sense https://github.com/RDycore/RDycore/compare/main...bishtgautam/cmake-fix

2. Following is the Ceed-related error

```
srun -n 1 -c 128 --cpu_bind=cores -G 1 --gpu-bind=single:1 ../../../bin/rdycore ex2b_dirichlet_bc.yaml -ceed /gpu/cuda -dm_vec_type cuda -use_gpu_aware_mpi 1 -options_left
[...]

PR #383, opened 2026-04-23
Adds upwind Roe Riemann solver for tracers

bishtgautam:
- This PR implements the upwind Roe Riemann solver that Dongyu implemented in his branch.
- The upwind Roe Riemann solver is implemented for PETSc and CEED version
- An MMS test for the upwind solver is added
   - The test is based on the existing MMS solver with default Roe Riemann solver
   - The L2 error tolerance for the upwind Roe solver is slightly lower (0.92) than for the default solver (0.93).

### DIFF
--- a/src/yaml_input.c
+++ b/src/yaml_input.c
@@ -122,8 +122,9 @@ static const cyaml_strval_t numerics_temporal_types[] = {
 
 // mapping of strings to numerics riemann solver types
 static const cyaml_strval_t numerics_riemann_types[] = {
-    {"roe",  RIEMANN_ROE },
-    {"hllc", RIEMANN_HLLC},
+    {"roe",           RIEMANN_ROE },
+    {"upwinded_roe",  RIEMANN_UPWINDED_ROE },
+    {"hllc",          RIEMANN_HLLC},
 };
 
 // mapping of numerics fields to members of RDyNumericsSection
@@ -926,8 +927,8 @@ static PetscErrorCode ValidateConfig(MPI_Comm comm, RDyConfig *config, PetscBool
   if (config->numerics.temporal == TEMPORAL_BEULER) {
     PetscCheck(PETSC_FALSE, comm, PETSC_ERR_USER, "The backward euler temporal method (BEULER) is not implemented.");
   }
-  if (config->numerics.riemann != RIEMANN_ROE) {
-    PetscCheck(PETSC_FALSE, comm, PETSC_ERR_USER, "Only the roe riemann solver (ROE) is currently implemented.");
+  if (config->numerics.riemann != RIEMANN_ROE && config->numerics.riemann != RIEMANN_UPWINDED_ROE) {
+    PetscCheck(PETSC_FALSE, comm, PETSC_ERR_USER, "Only the roe riemann solver (ROE) and upwinded roe riemann solver (UPWINDED_ROE) are currently implemented.");
   }
 
   PetscCheck(strlen(config->grid.file), comm, PETSC_ERR_USER, "grid.file not specified!");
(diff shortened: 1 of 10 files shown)


src/yaml_input.c
```diff
@@ -926,8 +927,9 @@ static PetscErrorCode ValidateConfig(MPI_Comm comm, RDyConfig *config, PetscBool
   if (config->numerics.temporal == TEMPORAL_BEULER) {
     PetscCheck(PETSC_FALSE, comm, PETSC_ERR_USER, "The backward euler temporal method (BEULER) is not implemented.");
   }
-  if (config->numerics.riemann != RIEMANN_ROE) {
-    PetscCheck(PETSC_FALSE, comm, PETSC_ERR_USER, "Only the roe riemann solver (ROE) is currently implemented.");
+  if (config->numerics.riemann != RIEMANN_ROE && config->numerics.riemann != RIEMANN_UPWIND_ROE) {
```
jeff-cohere, line 930: Since only one of these solver is valid for the SWE-only system, should we reject the case where `config->physics.sediment.num_classes == 0 && config->numerics.riemann != RIEMANN_ROE`?

jeff-cohere: Also, down the line, if this upwinded tracer method is demonstrably better than the not-upwinded case, we could interpret `RIEMANN_ROE` as an upwinded Roe solver for the tracers. But I must admit I'm not familiar with cases where you wouldn't want upwinding for the tracers.

jeff-cohere (approved):
Looks good to me. If we start adding more Riemann solvers, we might want to consider carefully which configurations we want to support, since it could turn into a lot of code in the approach we're currently following.

fdongyu (approved):
Looks good to me!

src/yaml_input.c
bishtgautam: I have now added a check during the setup of the operator to report an error if anything other than RIEMANN_ROE is being used for flow-only case.

bishtgautam:
@copilot after the merge of this PR, I'm getting following error when running on Perlmutter GPUs

```
/global/cfs/cdirs/m4267/petsc/petsc_v3.24.1/pm-gpu-hdf5_1_14_3-kokkos_4_6_00-opt-32bit-gcc-13-2-1-b522cb8c110/externalpackages/libCEED-main/interface/ceed-vector.c:608 in CeedVectorGetArrayWrite(): Cannot grant CeedVector array access, a process has read access
```

The relevant settings in the YAML file are as follows
```
physics:
  flow:
    mode: swe
    h_anuga_reg_parameter: 1.e-3

numerics:
  spatial: fv
  temporal: euler
  riemann: roe
```

Do you see any obvious thing that is incorrect in this PR?

jeff-cohere:
This is the error message I was seeing when a CEED header was placed outside of the directory in which its corresponding source lives. It's a "race condition" that results from a problem in the fallback code that executes after the JIT compiler fails. See #373 

(IF it's the same issue)

bishtgautam:
@jeff-cohere I looked at the https://github.com/RDycore/RDycore/pull/376, which fixed the issue. But I'm not sure what needs to be fixed here.

jeff-cohere:
Do you have a traceback of where this error occurs? That might tell us which JIT compile is failing.

bishtgautam:
Here is the log https://gist.githubusercontent.com/bishtgautam/98c7e033045c9c4d0eba66216bef715e/raw/795196e947195bb43cd7cdb3fd870c85e0e4dc13/gistfile1.txt

I'm running an optimized version of PETSc. I could install the debugging version of PETSc and see if I get more helpful information.

PR #387, opened 2026-04-29
Adds forcing module (Phase A refactor)

bishtgautam:
Moves dataset I/O and forcing logic from driver/main.c into a new internal module src/forcing/:

- src/forcing/rdyforcing.c: lifecycle (Create/Setup/Apply/Destroy), CLI option parsing, and dispatch to RDycore source/BC setters
- src/forcing/rdyforcing_dataset.c: binary PETSc Vec I/O for constant, homogeneous, raster, and unstructured datasets
- src/forcing/rdyforcing_map.c: nearest-neighbor spatial mapping between dataset grids and the RDycore mesh
- include/private/rdyforcingimpl.h: internal types and declarations

RDyForcing is embedded in struct _p_RDy (rdycoreimpl.h) and its lifecycle is managed automatically by RDySetup, RDyAdvance, and RDyDestroy. No public API is exposed. driver/main.c is simplified to a plain create/setup/advance/destroy loop.


src/forcing/rdyforcing_dataset.c
```diff
@@ -0,0 +1,391 @@
...
+
+  PetscViewer viewer;
+  PetscCall(VecCreate(PETSC_COMM_SELF, &data->data_vec));
+  PetscCall(PetscViewerBinaryOpen(PETSC_COMM_SELF, data->file, FILE_MODE_READ, &viewer));
+  PetscCall(VecLoad(data->data_vec, viewer));
+  PetscCall(PetscViewerDestroy(&viewer));
+
+  PetscInt size;
+  PetscCall(VecGetSize(data->data_vec, &size));
+  PetscCall(VecGetArray(data->data_vec, &data->data_ptr));
```
jeff-cohere: Getting a pointer to the Vec for the entire duration of the dataset's lifetime might be fine, but it's usually easier to reason about the state of the pointer when it exists only when used. For example, you might call this (and `VecRestoreArray`) before and after reading the data as in e.g. `RDyForcingSetRasterData`.

This prevents any intermediate operations from messing with the validity of the pointer.

jeff-cohere: This comment applies to all of the cases in which you're calling `VecGetArray` and `VecRestoreArray` in separate functions from where the data is accessed. I know we do this in some of the operator logic, but in that case the pointers have a lifetime that is explicitly defined by that of `OperatorData`, which is a bit easier to understand.

bishtgautam: @jeff-cohere, I believe I understand your comment, but let me verify that I understood it correctly. So, the code here could be modified as:

```c
  PetscInt size;
  PetscCall(VecGetSize(data->data_vec, &size));
  PetscCall(VecGetArray(data->data_vec, &data->data_ptr));

  data->ndata  = data->data_ptr[0];
  data->stride = data->data_ptr[1];
  PetscCall(VecResotoreArray(data->data_vec, &data->data_ptr));

``` 

and the `RDyForcingSetUnstructuredData()` can be modified as
```c
PetscErrorCode RDyForcingSetUnstructuredData(...) {

  ...
  ...

  // NEW code
  PetscCall(VecGetArray(data->data_vec, &data->data_ptr));

  PetscInt offset = 2;
  PetscInt stride = data->stride;

  for (PetscInt icell = 0; icell < data->mesh_nelements; icell++) {
    PetscInt idx = data->data2mesh_idx[icell] * stride;

    for (PetscInt ii = 0; ii < stride; ii++) {
      data_values[icell * stride + ii] = data->data_ptr[idx + ii + offset];
    }
  }

  // NEW code
  PetscCall(VecResotoreArray(data->data_vec, &data->data_ptr));

}
```

Did I get it?

bishtgautam: In https://github.com/RDycore/RDycore/pull/387/changes/40bf2bb93b757cfdd2fb03881d6729870389ad75, I got rid of `PetscScalar *vec_ptr;` from all the structs and instead am using local variables to get access to the Vec.

jeff-cohere (approved):

bishtgautam:
The last commit was mistakenly missed when this branch was merged to main. The #396 adds the missing commit.

PR #392, opened 2026-05-09
Adds support to output time-average quantities

### LINKED ISSUE #391: Output temporally average quantities
Allow RDycore to output temporally average quantities. For SWE:
- h
- hu
- hv
- u
- v
- Sum(Mass_Flux_out_of_cell)

bishtgautam:
We can now output:
- Primitive variables: `VelocityX`, `VelocityY`, `SedimentMassPerUnitArea0`.
- Time-average variables by adding a suffix `_Mean` to their names.

An example of the output block is as follows:

```
output:
  format: xdmf
  output_interval: 100
  fields:
    - Height
    - MomentumX
    - MomentumY
    - WaterSource
    - MomentumXSource
    - Height_Mean
    - MomentumY_Mean
    - MomentumXSource_Mean
    - VelocityX
    - VelocityY
    - VelocityX_Mean
    - VelocityY_Mean
```

Fixes #391


include/private/rdycoreimpl.h
```diff
@@ -137,10 +137,32 @@ struct _p_RDy {
...
-  DM               dm_diags;
-  SectionFieldSpec field_diags;
-  Vec              vec_diags;
+  PetscInt last_accumulated_step;  // last step for variables were accumulated for time-averaged output
+
+  // field spec, Vec, accumulator, and accumulated time for time-averaged solution output
+  SectionFieldSpec soln_avg_fields;  // Height_Mean, MomentumX_Mean, MomentumY_Mean, SedimentMassPerUnitAreaN_Mean
+  Vec              vec_soln_avg;     // time-averaged solution (same layout as u_global)
+  Vec              vec_soln_accum;   // running sum for time averaging
+  PetscReal        soln_accumulated_time;
```
jeff-cohere: Do we expect there to be distinct averaging time windows for our solution, primitive variables, and sources?

include/private/rdyoperatorimpl.h
```diff
@@ -144,6 +144,15 @@ typedef struct Operator {
...
       // domain-wide flux_divergence vector;
       CeedVector flux_divergence;
+
+      // domain-wide primitive variables vector (current step, output from source Q-function)
+      CeedVector primitive_variables;
+      // domain-wide primitive variables accumulation vector (dt-weighted sum)
+      CeedVector primitive_variables_accum;
+
+      // instantaneous external source snapshot and accumulation for time-averaged source output
+      CeedVector ceed_src_inst;   // instantaneous source snapshot
```
jeff-cohere, line 162: We probably don't need the `ceed_` prefix for these fields, given that they are inside a `ceed` struct.

src/rdysetup.c
```diff
@@ -1414,38 +1414,72 @@ PetscErrorCode RDySetup(RDy rdy) {
...
-          rdy->field_diags.num_fields = 1;
-          strcpy(rdy->field_diags.field_names[0], "Diagnostics");
-          rdy->field_diags.num_field_components[0] = 0;
-        }
-        strcpy(rdy->field_diags.field_component_names[0][rdy->field_diags.num_field_components[0]], rdy->config.output.fields[i]);
-        ++rdy->field_diags.num_field_components[0];
-      }
+  // set up source output field specs (instantaneous and time-averaged)
+  {
+    PetscInt num_src_comp                        = 3 + rdy->config.physics.sediment.num_classes;
```
jeff-cohere: We don't support it yet, but when we do add heat transfer and salinity, we'll have to adjust this logic.

src/xdmf_output.c
```diff
@@ -1,92 +1,180 @@
...
+                               "        </DataItem>\n"
+                               "      </Attribute>\n",
+                               name, num_cells_global, h5_basename, group_name, name));
+      }
+    }
+  }
   PetscFunctionReturn(PETSC_SUCCESS);
 }
 
+// writes output fields that match the given set; writes no fields when fields_count == 0
```
jeff-cohere, line 141: What is this comment describing?

include/private/rdycoreimpl.h
jeff-cohere: If not, we could use a single time accumulation variable as with `last_accumulated_step`.

include/private/rdycoreimpl.h
```diff
@@ -137,10 +137,32 @@ struct _p_RDy {
...
+
+  // field spec, Vec, and accumulated time for time-averaged primitive variables output
+  // (uses the same DM layout as dm above)
+  SectionFieldSpec prim_vars_fields;
```
jeff-cohere: I can't help but wonder if there's some way to reduce the amount of code related to all these quantities we were formerly calling "diagnostics". It's true that there are several distinct cases for how to compute and update these diagnostics, so the explicitness of the current approach makes some sense, and this is definitely the type of code that an LLM would write (or suggest). But to me it's pretty noisy when taken in with with the rest of the code--one has to know which functions to call in what order on which diagnostic, and what assumptions all those functions make.

I don't know how much it matters, but I can probably think of an "old world" way to organize this stuff if we want to reduce the noise level.

include/private/rdyoperatorimpl.h
```diff
@@ -236,4 +254,7 @@ PETSC_INTERN PetscErrorCode ResetOperatorDiagnostics(Operator *);
 PETSC_INTERN PetscErrorCode UpdateOperatorDiagnostics(Operator *);
 PETSC_INTERN PetscErrorCode GetOperatorDiagnostics(Operator *, OperatorDiagnostics *);
 
+// source output
+PETSC_INTERN PetscErrorCode AddOperatorSourceVariables(Operator *);
```
jeff-cohere, line 273: Do we need to expose this function, or can we make it `static` and local to `operator.c` like `AddOperatorPrimitiveVariables`?

jeff-cohere (approved):
This is a lot of new code, and it seems to do what is intended, but things are starting to feel a little complicated to me. It could be reorganized. Of course my design disposition is now somewhat old-fashioned, and I think it's worth discussing how much we should strive for simplicity in an age where (for example) one can ask Claude to help with adding new features.

include/private/rdycoreimpl.h
bishtgautam: This is a good size PR to experiment with LLM to organize code the "old world" way. It would be a good learning case study. Let's have a discussion next week on how to refactor this PR.

PR #393, opened 2026-05-22
2r petsc

wisangsugiarta:
Added 2R for CPU only using Petsc. Achieves identical MMS convergence as Hybrid.


src/swe/swe_petsc.c
```diff
@@ -115,20 +122,41 @@ static PetscErrorCode ApplyInteriorFlux(void *context, PetscOperatorFields field
...
   PetscReal        *flux_vec_int = data_edge->fluxes;
 
-  // Collect the h/hu/hv for left and right cells to compute u/v
+  // MUSCL: compute cell-centered gradients and reconstruct face states
+  if (interior_flux_op->use_slope_reconstruction) {
+    PetscCall(ComputeLeastSquaresGradients(mesh, interior_flux_op->ls_grad_coeffs, u_ptr, interior_flux_op->grad_h, interior_flux_op->grad_hu,
+                                           interior_flux_op->grad_hv));
+    PetscCall(ReconstructFaceValues(mesh, u_ptr, interior_flux_op->grad_h, interior_flux_op->grad_hu, interior_flux_op->grad_hv,
+                                    interior_flux_op->use_limiter, interior_flux_op->q_reconstructed));
+  }
```
bishtgautam: Instead of putting all the new code within `if (interior_flux_op->use_slope_reconstruction)`, it might be a better approach to create a new function `ApplyInteriorFluxUsingSlopeRecon` something similar to `ApplyInteriorFluxHR`.

Maybe @jeff-cohere also has some suggestions.

src/swe/swe_petsc.c
```diff
@@ -172,14 +222,48 @@ static PetscErrorCode ApplyInteriorFlux(void *context, PetscOperatorFields field
...
-            PetscInt left_owned_cell_id = cells->local_to_owned[left_local_cell_id];
-            f_ptr[n_dof * left_owned_cell_id + i_dof] += flux_vec_int[n_dof * e + i_dof] * (-edge_len / areal);
+          rhs_local_ptr[n_dof * left_local_cell_id + i_dof] += flux_vec_int[n_dof * e + i_dof] * (-edge_len / areal);
+          rhs_local_ptr[n_dof * right_local_cell_id + i_dof] += flux_vec_int[n_dof * e + i_dof] * (edge_len / arear);
+        }
+      }
+    }
+
+    PetscCall(VecRestoreArray(rhs_local, &rhs_local_ptr));
+    PetscCall(DMLocalToGlobal(dm, rhs_local, ADD_VALUES, f_global));
```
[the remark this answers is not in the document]
wisangsugiarta: Resolved! Was a true mistake

src/swe/swe_petsc.c
wisangsugiarta: Thanks Gautam, I fixed this by creating a new function ApplyInteriorFlux2R, it makes it much more readable.

src/swe/swe_petsc.c
```diff
@@ -84,21 +85,156 @@ typedef struct {
...
+static PetscErrorCode ApplyInteriorFlux(void *context, PetscOperatorFields fields, PetscReal dt, Vec u_local, Vec f_global) {
+  PetscFunctionBegin;
+
   InteriorFluxOperator *interior_flux_op = context;
 
+  // dispatch to MUSCL second-order path if enabled
+  if (interior_flux_op->use_slope_reconstruction) {
+    PetscCall(ApplyInteriorFlux2R(interior_flux_op, dt, u_local, f_global));
+    PetscFunctionReturn(PETSC_SUCCESS);
+  }
```
bishtgautam: @wisangsugiarta, The change here is the correct direction but let's take it one step further such that we register `ApplyInteriorFlux2R` when `interior_flux_op->use_slope_reconstruction` is turned on. To do so, you will need to add a `if`-statement around https://github.com/wisangsugiarta/RDycore/blob/2r_petsc/src/operator.c#L195-L216

```
if (interior_flux_op->use_slope_reconstruction) {
  // new code
  PetscCall(CreatePetscFlux2ROperator(...))
  ...
} else {
   // existing code
   switch (op->config->physics.flow.well_balancing) 
   ...
   ...
   ...
}
```

How does this sound?

jeff-cohere: Alternatively, you can check `interior_flux_op->use_slope_reconstruction` and pass `ApplyInteriorFlux2R` as the `Apply` function when creating the interior flux operator in `CreatePetscSWEInteriorFluxOperator`.

jeff-cohere (approved):

src/swe/swe_petsc.c
wisangsugiarta: You both are right, that makes a lot more sense! The new commit follows that design pattern.

src/swe/swe_petsc.c
```diff
@@ -84,21 +85,152 @@ typedef struct {
...
+
+  // compute cell-centered gradients and reconstruct face states
+  PetscCall(ComputeLeastSquaresGradients(mesh, interior_flux_op->ls_grad_coeffs, u_ptr, interior_flux_op->grad_h, interior_flux_op->grad_hu,
+                                         interior_flux_op->grad_hv));
+  PetscCall(ReconstructFaceValues(mesh, u_ptr, interior_flux_op->grad_h, interior_flux_op->grad_hu, interior_flux_op->grad_hv,
+                                  interior_flux_op->use_limiter, interior_flux_op->q_reconstructed));
+
+  // Collect the h/hu/hv for left and right cells to compute u/v.
+  // For owned interior edges, use reconstructed face states;
+  // for non-owned (ghost-side) edges, fall back to cell-averaged values
```
bishtgautam: @wisangsugiarta, For non-owned cells, why are we using cell-averages and not the reconstructed face states? The current approach will cause a problem because the flux through the face on the two MPI ranks could be computed differently.

wisangsugiarta: Fair concern, the cell-average branch for non-owned edges doesn't actually feed f_global — the accumulation loop skips non-owned edges and the owner's contribution is communicated back via DMLocalToGlobal(ADD_VALUES). So every interior face flux is computed exactly once by the edge owner, and shared.
There is a place where ghost data approaches 1st order. When rank A reconstructs the face state and the neighbor is its ghost cell, the ghost's gradient is incomplete. So rank A reconstructs the ghost side to first order. The flux is still consistent, but it's locally first-order-accurate at partition boundaries - but this is also the case with the ceed/hybrud. I have to be honest that most of my experiments are not in parallel which is why I didn't include but I looked into it and it would be pretty simple to add. I could change both to keep second order at partitions bit looking at the mms studies the difference was quite small. Let me know!

wisangsugiarta: I had time to fix, its now exactly the same for mpi > 1

src/swe/swe_petsc.c
```diff
@@ -84,21 +85,155 @@ typedef struct {
...
+    PetscInt  left_local_cell_id  = edges->cell_ids[2 * edge_id];
+    PetscInt  right_local_cell_id = edges->cell_ids[2 * edge_id + 1];
+    PetscBool owned               = edges->is_owned[edge_id];
+
+    if (right_local_cell_id != -1) {
+      if (owned) {
+        datal->h[e]  = fmax(0.0, interior_flux_op->q_reconstructed[owned_e * 6 + 0]);
+        datal->hu[e] = interior_flux_op->q_reconstructed[owned_e * 6 + 1];
+        datal->hv[e] = interior_flux_op->q_reconstructed[owned_e * 6 + 2];
+        datar->h[e]  = fmax(0.0, interior_flux_op->q_reconstructed[owned_e * 6 + 3]);
```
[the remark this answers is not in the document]
wisangsugiarta: Would only matter for complex numbers which we dont support.

src/operator_fluxes_ceed.c
```diff
@@ -1038,6 +1038,65 @@ PetscErrorCode ComputeLeastSquaresGradients(RDyMesh *mesh, const PetscReal *ls_g
...
+    PetscCall(VecGetArray(g_global, &g_ptr));
+    for (PetscInt c = 0; c < mesh->num_cells; c++) {
+      if (cells->is_owned[c]) {
+        PetscInt owned       = cells->local_to_owned[c];
+        g_ptr[3 * owned + 0] = grads[0][c * 2 + k];
+        g_ptr[3 * owned + 1] = grads[1][c * 2 + k];
+        g_ptr[3 * owned + 2] = grads[2][c * 2 + k];
+      }
+    }
+    PetscCall(VecRestoreArray(g_global, &g_ptr));
```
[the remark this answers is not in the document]
wisangsugiarta: Im not sure this is actually needed but seems fair to zero out before any pass, added.

src/swe/swe_petsc.c
bishtgautam: @wisangsugiarta, I believe I understand how this is working in parallel, but I have a few additional comments.

src/swe/swe_petsc.c
```diff
@@ -84,21 +85,155 @@ typedef struct {
...
+        datar->hv[e] = interior_flux_op->q_reconstructed[owned_e * 6 + 5];
+        owned_e++;
+      } else {
+        datal->h[e]  = u_ptr[n_dof * left_local_cell_id + 0];
+        datal->hu[e] = u_ptr[n_dof * left_local_cell_id + 1];
+        datal->hv[e] = u_ptr[n_dof * left_local_cell_id + 2];
+        datar->h[e]  = u_ptr[n_dof * right_local_cell_id + 0];
+        datar->hu[e] = u_ptr[n_dof * right_local_cell_id + 1];
+        datar->hv[e] = u_ptr[n_dof * right_local_cell_id + 2];
+      }
```
bishtgautam: For non-owned edges, I believe these a just placeholder values in `datal` and `datar` that are send to Riemann solver. However, the flux computed for these never used. We should add a comment here describing that this is what is happening here.

src/swe/swe_petsc.c
```diff
@@ -84,21 +85,155 @@ typedef struct {
...
+        rhs_local_ptr[n_dof * left_local_cell_id + i_dof] += flux_vec_int[n_dof * e + i_dof] * (-edge_len / areal);
+        rhs_local_ptr[n_dof * right_local_cell_id + i_dof] += flux_vec_int[n_dof * e + i_dof] * (edge_len / arear);
+      }
+    }
+  }
+
+  PetscCall(VecRestoreArray(rhs_local, &rhs_local_ptr));
+  PetscCall(VecRestoreArray(u_local, &u_ptr));
+  PetscCall(DMLocalToGlobal(dm, rhs_local, ADD_VALUES, f_global));
+  PetscCall(DMRestoreLocalVector(dm, &rhs_local));
```
bishtgautam, line 210: In the CEED-version of the code, the `DMLocalToGlobal` is called after fluxes across internal **AND** boundary edges is computed. We should add a comment that notes:
1. Difference between the `DMLocalToGlobal` being called differently for PETSc and CEED versions, and
2. That `DMLocalToGlobal` is needed for 2R, but not for the first order.

src/swe/swe_petsc.c
wisangsugiarta: Yes! for non-owned edges the datal/datar are just placeholders to keep the batched solve well-defined, and that flux is skipped as its a non-owned edge. Each edge's real contribution is computed once by its owning rank and added back via DMLocalToGlobal. Added a comment in the code

src/swe/swe_petsc.c
bishtgautam: Thanks for the comment.

bishtgautam (approved):
@wisangsugiarta, Thanks for addressing the comments and your patience in getting this approved.

PR #395, opened 2026-06-05
Fixes the computation of slope for quad cells

bishtgautam:
Corrects the computation of slope for quad cells.


yixiao2 (approved):
Looks good to me, thanks for the fix!

